PR #131, opened 2026-05-30
fix: parameterize node_id lookup to prevent SQL injection in webhook handlers

Joshua-Medvinsky:
## Problem

`backend/app/api/slack.py:77` and `backend/app/api/telegram.py:73` look up the workflow for an incoming webhook by manually constructing a raw SQL fragment that interpolates the caller-supplied `node_id` path parameter into a JSON literal via f-string. The existing code strips single and double quotes from `node_id` before interpolation, so a clean SQL-string breakout could not be validated. However, hand-building SQL fragments this way is fragile and not idiomatic.

## Why change it (defense-in-depth / input robustness)

**Location:** `backend/app/api/slack.py:77`, `backend/app/api/telegram.py:73`

Both handlers used:

```python
query = text(f"SELECT ... WHERE nodes::jsonb @> '[{{\"id\": \"{node_id}\"}}]'::jsonb")
```

Relying on manual sanitization (stripping quotes) to build a SQL/JSON fragment is brittle:

- It couples query correctness to an ad-hoc character-stripping step that is easy to break in future edits.
- Inputs that are not quote-related — e.g. backslashes or other escape sequences — can still produce invalid JSON or runtime errors rather than a clean lookup.
- It is harder to read and reason about than a bound parameter.

This is a defense-in-depth / robustness improvement, not a fix for a confirmed exploitable injection. Parameter binding removes the fragile manual sanitization and lets the database driver handle escaping.

## Fix

Replace the f-string interpolation with a proper parameterized query using SQLAlchemy's `bindparams`:

```python
text("nodes::jsonb @> (:node_filter)::jsonb").bindparams(
    node_filter=json.dumps([{"id": node_id}])
)
```

`json.dumps` produces correctly-escaped JSON, and the bind parameter ensures the value is never interpolated into the SQL string. The parenthesized cast `(:node_filter)::jsonb` is required because SQLAlchemy's bind-param parser would otherwise consume the trailing `::jsonb` as part of the parameter name.

Both the Slack and Telegram handlers now share the same safe, idiomatic pattern.

## Steps to reproduce (behavior before fix)

1. Send a webhook POST to the Slack or Telegram handler with a crafted `node_id` path parameter containing characters not handled by the quote-stripping (e.g. backslashes / escape sequences).
2. Observe that the manually-built JSON fragment can become malformed, producing invalid JSON / query errors rather than a clean lookup.
3. With the parameterized query, the same input is safely encoded as a JSON string value and the lookup behaves correctly.

## Test Plan

- [x] Parenthesized cast `(:node_filter)` appears in compiled SQL for both Slack and Telegram handlers
- [x] The `node_filter` bind value is valid JSON matching `[{"id": "<node_id>"}]`
- [x] Special characters in `node_id` (quotes, semicolons, SQL keywords, backslashes) are safely encoded inside the JSON string value
- [x] `ruff check` and `ruff format` pass
- [x] Tests use `IsolatedAsyncioTestCase` (compatible with Python ≥ 3.14)

### DIFF
--- a/backend/app/api/slack.py
+++ b/backend/app/api/slack.py
@@ -3,11 +3,10 @@
 Handles URL verification challenge and routes events to workflow execution.
 """
 
-import json
-
 import asyncio
 import hashlib
 import hmac
+import json
 import logging
 import time
 import uuid
@@ -82,7 +81,7 @@ async def _find_workflow_by_node_id(
     """Use JSONB containment to find the workflow containing this node_id."""
     result = await db.execute(
         select(Workflow).where(
-            text("nodes::jsonb @> :node_filter::jsonb").bindparams(
+            text("nodes::jsonb @> (:node_filter)::jsonb").bindparams(
                 node_filter=json.dumps([{"id": node_id}])
             )
         )
--- a/backend/app/api/slack.py
+++ b/backend/app/api/slack.py
@@ -3,6 +3,8 @@
 Handles URL verification challenge and routes events to workflow execution.
 """
 
+import json
+
 import asyncio
 import hashlib
 import hmac
@@ -78,9 +80,12 @@ async def _find_workflow_by_node_id(
     node_id: str,
 ) -> Workflow | None:
     """Use JSONB containment to find the workflow containing this node_id."""
-    safe_node_id = node_id.replace("'", "").replace('"', "")
     result = await db.execute(
-        select(Workflow).where(text(f'nodes::jsonb @> \'[{{"id": "{safe_node_id}"}}]\'::jsonb'))
+        select(Workflow).where(
+            text("nodes::jsonb @> :node_filter::jsonb").bindparams(
+                node_filter=json.dumps([{"id": node_id}])
+            )
+        )
     )
     return result.scalar_one_or_none()
 
--- a/backend/app/api/telegram.py
+++ b/backend/app/api/telegram.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import hmac
+import json
 import logging
 import uuid
 from datetime import datetime, timezone
@@ -74,9 +75,12 @@ async def _find_workflow_by_node_id(
     node_id: str,
 ) -> Workflow | None:
     """Use JSONB containment to find the workflow containing this node_id."""
-    safe_node_id = node_id.replace("'", "").replace('"', "")
     result = await db.execute(
-        select(Workflow).where(text(f'nodes::jsonb @> \'[{{"id": "{safe_node_id}"}}]\'::jsonb'))
+        select(Workflow).where(
+            text("nodes::jsonb @> :node_filter::jsonb").bindparams(
+                node_filter=json.dumps([{"id": node_id}])
+            )
+        )
     )
     return result.scalar_one_or_none()
 
--- a/backend/app/api/telegram.py
+++ b/backend/app/api/telegram.py
@@ -77,7 +77,7 @@ async def _find_workflow_by_node_id(
     """Use JSONB containment to find the workflow containing this node_id."""
     result = await db.execute(
         select(Workflow).where(
-            text("nodes::jsonb @> :node_filter::jsonb").bindparams(
+            text("nodes::jsonb @> (:node_filter)::jsonb").bindparams(
                 node_filter=json.dumps([{"id": node_id}])
             )
         )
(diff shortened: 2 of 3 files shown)


mbakgun:
Hi Joshua, thanks for taking the time to investigate this and put together a fix. We @ckakgun  appreciate the security focus and the detailed write-up.

After reviewing the PR, we're going to hold off on merging it in its current form.

There are a few issues with the proposed implementation:

- `:node_filter::jsonb` does not parse as intended inside SQLAlchemy's `text()` construct. The bind-parameter matcher reads the name as `node_filte`, causing `_find_workflow_by_node_id` to raise `ArgumentError` before the query is executed. Using `(:node_filter)::jsonb` or `CAST(:node_filter AS jsonb)` resolves this.

- The new `json` import in `slack.py` is in a separate block and fails the project's Ruff import-order check.

We also weren't able to validate the high-severity SQL injection scenario described in the PR. The current implementation removes both single and double quotes before interpolation, so an `OR 1=1`-style payload remains inside the JSON value rather than escaping into SQL. Also, `node_id` comes from the webhook URL path, not the event payload.

That said, the underlying point is still valid: manually constructing SQL fragments is brittle. Malformed values such as backslashes can produce invalid JSON and result in errors, so moving to a properly parameterized query is the right direction as a defense-in-depth and input-robustness improvement.

We'd be happy to take another look if you'd like to revise the PR with one of the casting fixes above, sort the import so Ruff passes, and add a small test that exercises `_find_workflow_by_node_id` without mocking it.

Thanks again for the contribution and for flagging this area.

Joshua-Medvinsky:
Thanks for the detailed review @mbakgun @ckakgun! I've addressed all three points:

1. **Casting bug fixed**: Changed `:node_filter::jsonb` to `(:node_filter)::jsonb` in both `slack.py` and `telegram.py`. This prevents SQLAlchemy from misreading the bind param name as `node_filte`.

2. **Import order fixed**: Merged the `json` import into the alphabetical stdlib block in `slack.py` so Ruff's I001 check passes.

3. **Test for _find_workflow_by_node_id**: Added `test_find_workflow_by_node_id.py` that exercises the parameterized query, validates the parenthesized cast, and confirms special characters in node_id are safely escaped inside the JSON value.

I also acknowledge the severity re-scoping — the PR is better framed as defense-in-depth and input robustness rather than a direct SQL injection, given the quote stripping and URL-path source. Happy to update the PR description if you'd like.

mbakgun:
Thanks @Joshua-Medvinsky, the production-code changes look good 🤗. I verified that the parenthesized cast now parses correctly in SQLAlchemy and the Slack import order is fixed.

There are still a few issues in the new test file before we can merge:

- `ruff check .` fails because `uuid`, `Mock`, `patch`, and `text` are imported but unused.
- `ruff format --check .` reports that the test file needs formatting.
- Please avoid `asyncio.get_event_loop().run_until_complete(...)`. This fails on Python 3.14, which is included by the project's `>=3.11` Python requirement. Using `IsolatedAsyncioTestCase` with `await` would fit the existing test style.
- The cast assertion is currently too permissive because the `or "node_filter" in compiled` branch can pass without proving that the parenthesized cast is present.
- Since both handlers changed, please exercise the Telegram lookup as well.
- Could you also update the PR description to reflect the defense-in-depth / input-robustness framing?

Thanks again. The implementation is close; it just needs the test cleanup.

mbakgun:
Thanks @Joshua-Medvinsky.  One small request regarding the PR description: could you please reframe this as a defense-in-depth and input-robustness improvement rather than a confirmed high-severity SQL injection vulnerability?

The previous implementation did interpolate into raw SQL, which is brittle and worth fixing, but it also removed both single and double quotes before building the query. Because of that, we were not able to validate the SQL breakout scenario described in the current text.

It would be more accurate to remove the `High` severity / confirmed CWE-89 framing and describe the change as replacing fragile manual sanitization with proper parameter binding.

The CI run is currently failing because of one remaining unused import: `from sqlalchemy import text`

Thanks again. Once those two small cleanups are in, we can take another look.

ckakgun:
Merged. Thanks for the contribution, @Joshua-Medvinsky 👑 , and for staying with it through the review rounds.

The final result is a clean improvement: parameter binding in both the Slack and Telegram handlers, and new test coverage for _find_workflow_by_node_id where we had none before. I appreciate how quickly you turned around the casting fix, the lint cleanup, and the test rewrite.

Thanks again, and feel free to open more PRs anytime.

PR #147, opened 2026-06-05
fix(web): fix password length & error info

eryue0220:
## Summary

The registration page was updated so client-side password rules match the backend (8 characters minimum) and API validation errors are shown more clearly to users.

## Changes
1. Password minimum length (6 → 8)
2. Improved API error message handling


mbakgun (approved):
Thanks

PR #148, opened 2026-06-05
feat: add generic webhook trigger node

bvasystems:
## Summary
  - Add a generic Webhook Trigger node with a node-specific `/api/webhooks/{node_id}` endpoint
  - Reuse existing workflow auth, rate limiting, caching, execution history, and response behavior
  - Expose request body, headers, query, method, and timestamp to downstream nodes
  - Add docs and backend tests

  ## Validation
  - Ran `./check.sh` with test `SECRET_KEY` and `ENCRYPTION_KEY`
  - 901 tests passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,7 +389,7 @@ cp .env.example .env
 
 | Category | Nodes |
 |----------|-------|
-| **Triggers** | Input (Webhook), Cron, RabbitMQ Receive, Error Handler |
+| **Triggers** | Input, Webhook Trigger, Cron, RabbitMQ Receive, Error Handler |
 | **AI** | LLM, AI Agent, Qdrant RAG |
 | **Logic** | Condition, Switch, Loop, Merge |
 | **Data** | Set, Variable, DataTable, Execute (sub-workflow) |
--- a/backend/app/api/webhooks.py
+++ b/backend/app/api/webhooks.py
@@ -1,0 +1,66 @@
+"""Generic HTTP webhook trigger endpoint."""
+
+import json
+import uuid
+
+from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, Request, status
+from sqlalchemy import select, text
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from app.api.deps import get_current_user_optional
+from app.api.workflows import execute_workflow_request
+from app.db.models import User, Workflow
+from app.db.session import get_db
+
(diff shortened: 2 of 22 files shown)


mbakgun:
Hi @bvasystems, thank you so much for taking the time to put this together. The implementation is thoughtful, and we appreciate the tests, docs, and care you put into the PR.

After reviewing it, we do not think we should merge this one in its current shape. The main reason is that this capability is already covered by the existing Input node and workflow execute endpoint.

The Input node is already the HTTP webhook entry point for a workflow. Workflows can be called through `POST /api/workflows/{workflow_id}/execute`, with workflow auth, rate limiting, response caching, execution history, and response handling already wired in. Incoming request data is already available downstream through the Input node label, for example `$userInput.body`, `$userInput.headers`, and `$userInput.query`.

Screenshot 2026-06-05 at 7 49 53 PM

For the external app webhook use case, where an app sends its own raw JSON payload, the existing generic webhook body mode already passes that payload through without reshaping it. The editor also already has the cURL dialog in the right side panel, which covers copying and testing the HTTP request from outside Heym.

Screenshot 2026-06-05 at 7 49 46 PM

Given that, adding a second `webhookTrigger` node type and a separate `/api/webhooks/{node_id}` public endpoint would create a parallel mental model for something the Input node already handles. We would rather keep this part of the product centered on one clear entry point.

A couple of smaller notes:

The `method` and `triggered_at` fields are useful ideas, and those could make sense as a small standalone improvement to the existing Input webhook path.

The node ID lookup would also need more design work before becoming a public endpoint, since finding workflows by scanning `nodes::jsonb` on every request could become expensive without the right indexing and abuse protection.

For a change that introduces a new first class node type and a new public API surface, we would prefer to start with a GitHub Discussion first so we can align on the product and API design before implementation. If there is a specific gap the current Input node flow does not cover, please open a Discussion and we can dig into that together.

Thanks again for contributing and for engaging with the project. 🙏

bvasystems:
Hi, @mbakgun! Thank you for the careful analysis and detailed explanation.

That makes sense. I now understand that the Input node plus `POST /api/workflows/{workflow_id}/execute` is intended to be the canonical entry point for the HTTP webhook,

and that the generic webhook body mode already covers the raw JSON payload use case.

For future contributions, I will open a discussion on GitHub first so we can align the product/API design before implementation.

Thank you again for taking the time to analyze it.

PR #150, opened 2026-06-06
fix: derive auth cookie security from public origin

5p00kyy:
## Summary
- derive auth cookie Secure behavior from FRONTEND_URL before falling back to CORS origins
- allow insecure auth cookies only for explicit local HTTP development origins
- add coverage for HTTPS deployments where localhost appears first in CORS_ORIGINS

## Why
The previous check used `settings.cors_origins.startswith("http://localhost")`. In a deployment with `CORS_ORIGINS=http://localhost:4017,https://heym.example.com` and `FRONTEND_URL=https://heym.example.com`, auth cookies would be issued without the Secure flag even though the public app is HTTPS.

## Verification
- `cd frontend && bun run lint && bun run typecheck && bun run build`
- `cd backend && uv run ruff format --check . && uv run ruff check .`
- `SECRET_KEY=test-secret-key-for-tests-only-32-bytes ENCRYPTION_KEY=a0b1c2d3e4f5a0b1c2d3e4f5a0b1c2d3e4f5a0b1c2d3e4f5a0b1c2d3e4f5a0b1 ./check.sh`

### DIFF
--- a/backend/app/api/auth.py
+++ b/backend/app/api/auth.py
@@ -1,3 +1,5 @@
+from urllib.parse import urlparse
+
 from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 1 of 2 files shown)


mbakgun:
Thanks for the fix 🙂 @5p00kyy CI is green and the direction looks good, but I think we should close one security edge case before merge.

If `FRONTEND_URL` is unset or left at the localhost default while `CORS_ORIGINS` points to an HTTPS production origin, the new logic will still treat the frontend as local dev and issue insecure auth cookies. That is a security downgrade from the old behavior.

Can you make local/default FRONTEND_URL fall back to CORS origins, and add a regression test like:

**_should_use_secure_auth_cookies("http://localhost:4017", ["https://heym.example.com"]) is True**

After that this looks good to merge.

5p00kyy:
Good catch, thanks. I updated the helper so a local/default `FRONTEND_URL` now falls back to configured CORS origins when they are present, and added the requested regression case for `http://localhost:4017` plus an HTTPS CORS origin.

Local checks on the amended head `dde0eee`:
- `SECRET_KEY=... ENCRYPTION_KEY=... uv run pytest tests/test_auth_cookie_security.py tests/test_settings.py tests/test_settings_validation.py -q` -> 18 passed
- `uv run ruff check app/api/auth.py tests/test_auth_cookie_security.py`
- `uv run ruff format --check app/api/auth.py tests/test_auth_cookie_security.py`
- `git diff --check`

The GitHub run is back in `action_required` for fork approval.

mbakgun (approved):
Thank you ! 🤗

PR #151, opened 2026-06-06
chore: remove unused vite config file

eryue0220:
## Summary

* Remove duplicated vite config file
* update ts configuration file


ckakgun:
Thanks for the cleanup @eryue0220 💯 . I agree with the direction here: removing the generated `frontend/vite.config.js` and `frontend/vite.config.d.ts` files is the right call, and the `vite.config.ts` typing improvements are a nice cleanup too.

I would hold off on merging until the generated artifact cleanup is complete, though. I tested this in a clean temporary checkout by removing the existing `frontend/tsconfig*.tsbuildinfo` files and running `bun run build`. With the current `frontend/tsconfig.node.json`, the build can recreate `frontend/vite.config.js` and `frontend/vite.config.d.ts` in the frontend root, because the referenced node tsconfig still emits outputs for `vite.config.ts`.

Could you update the TypeScript config or build setup so these generated Vite config outputs are written only to an ignored location, or otherwise do not get emitted into the frontend root? For example, an ignored `outDir` for `tsconfig.node.json` would keep those generated files out of the project root.

One related cleanup: `frontend/tsconfig.tsbuildinfo` and `frontend/tsconfig.node.tsbuildinfo` are generated build cache files. `.gitignore` already ignores `*.tsbuildinfo`, but these files are currently tracked, so this PR updates generated state instead of removing it. Please remove them from version control as part of this cleanup, for example with `git rm --cached frontend/tsconfig.tsbuildinfo frontend/tsconfig.node.tsbuildinfo`.

After those changes, this should be good to merge. 🚀

eryue0220:
> Thanks for the cleanup @eryue0220 💯 . I agree with the direction here: removing the generated `frontend/vite.config.js` and `frontend/vite.config.d.ts` files is the right call, and the `vite.config.ts` typing improvements are a nice cleanup too.
> 
> I would hold off on merging until the generated artifact cleanup is complete, though. I tested this in a clean temporary checkout by removing the existing `frontend/tsconfig*.tsbuildinfo` files and running `bun run build`. With the current `frontend/tsconfig.node.json`, the build can recreate `frontend/vite.config.js` and `frontend/vite.config.d.ts` in the frontend root, because the referenced node tsconfig still emits outputs for `vite.config.ts`.
> 
> Could you update the TypeScript config or build setup so these generated Vite config outputs are written only to an ignored location, or otherwise do not get emitted into the frontend root? For example, an ignored `outDir` for `tsconfig.node.json` would keep those generated files out of the project root.
> 
> One related cleanup: `frontend/tsconfig.tsbuildinfo` and `frontend/tsconfig.node.tsbuildinfo` are generated build cache files. `.gitignore` already ignores `*.tsbuildinfo`, but these files are currently tracked, so this PR updates generated state instead of removing it. Please remove them from version control as part of this cleanup, for example with `git rm --cached frontend/tsconfig.tsbuildinfo frontend/tsconfig.node.tsbuildinfo`.
> 
> After those changes, this should be good to merge. 🚀

Got it, and it has been updated.

mbakgun:
Thanks !

PR #153, opened 2026-06-06
Harden Drive URL downloads

VykosMolt:
The Drive `downloadUrl` operation followed user-supplied URLs and used remote filenames to build storage paths. A download could reach an internal service, and a filename such as `../evil.txt` could escape the intended directory.

Downloads now require an HTTP or HTTPS URL that resolves to a globally routable address. The request connects to that validated address while preserving the original Host header and TLS server name. Checking DNS and then letting the HTTP client resolve the hostname again would leave a rebinding gap, so the connection uses the checked IP directly. Redirects are followed manually, with the same check before each new connection.

Downloaded and converted files also go through the existing filename and safe-storage-path helpers instead of constructing paths directly.

## Verification

The added tests cover unsafe remote filenames, non-public destinations, connecting to the validated address, and redirects to private addresses. Checks recorded for this change:

- `uv run pytest -q tests/test_drive_node.py tests/test_file_storage.py`
- `uv run ruff format --check app/services/workflow_executor.py app/config.py tests/test_drive_node.py`
- `uv run ruff check app/services/workflow_executor.py app/config.py tests/test_drive_node.py`
- `./run_tests.sh`

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -61,7 +61,6 @@ class Settings(BaseSettings):
     docs_dir: str = ""  # Path to docs content. Empty = use frontend/src/docs/content
     file_storage_dir: str = "./data/files"
     file_max_size_mb: int = 99
-    drive_download_allow_private_ips: bool = False
     mcp_protocol_max_concurrency: int = 20
     mcp_sse_max_sessions: int = 100
     app_version: str = ""
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -61,6 +61,7 @@ class Settings(BaseSettings):
     docs_dir: str = ""  # Path to docs content. Empty = use frontend/src/docs/content
     file_storage_dir: str = "./data/files"
     file_max_size_mb: int = 99
+    drive_download_allow_private_ips: bool = False
     mcp_protocol_max_concurrency: int = 20
     mcp_sse_max_sessions: int = 100
     app_version: str = ""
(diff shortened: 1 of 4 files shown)


mbakgun (changes requested):
Thanks for putting this together @VykosMolt  🙏 . The direction is right and the path traversal side looks solid to me. Reusing `_normalize_storage_filename` and `_safe_storage_path` for both downloaded and converted files is the right fix, and the manual redirect validation is a good improvement.

Before merging, I think we should tighten the SSRF protection. `_validate_drive_download_url()` resolves and checks the hostname, but `_fetch_drive_download_url()` then passes the original hostname to `httpx`, which resolves it again at connect time. That leaves a DNS rebinding / TOCTOU bypass: validation can see a public IP, while the actual connection resolves to a private, loopback, link-local, or metadata address.

I would also prefer this to be an allowlist based on `address.is_global` rather than a denylist of individual categories. The current checks can miss non-global ranges such as `100.64.0.0/10`, and this style is more fragile across IP classification edge cases.

Could we update the download path so the address validated is the address actually used for the connection, and only allow globally routable targets? The filename/path changes look good as is. Once the SSRF check is connect-time safe and global-only, this should be good to merge.

VykosMolt:
Thanks for the thorough review. Agreed on all of it. The DNS rebinding/TOCTOU gap is real, and pinning the connection to the validated address plus an alowlist is the right call over the denylist. I'll get to it and push an update.

VykosMolt:
Done!

mbakgun:
Thanks again for the update. I updated the PR description to remove the stale `DRIVE_DOWNLOAD_ALLOW_PRIVATE_IPS` note and align it with the final implementation.

The security concerns from my review are addressed now, CI is green, and this looks good to merge.

PR #154, opened 2026-06-06
Harden OAuth token handling

VykosMolt:
Public OAuth clients could start an authorization flow without PKCE, and issued access and refresh tokens were stored in plaintext. This change requires S256 PKCE for public clients at both authorization and token exchange, rejects unsupported client authentication methods, and uses a timing-safe comparison for the PKCE challenge.

New tokens are stored as SHA-256 hashes. Both MCP authentication paths can look up the new hashed rows as well as legacy plaintext rows, so existing sessions keep working during the transition. This PR does not migrate or revoke existing plaintext tokens; they remain usable until expiry or rotation.

Authorization-code and refresh-token grants also acquire row locks where the database supports them, to serialize the replay-sensitive lookups.

## Verification

The tests cover the authorization and token endpoints, public-client PKCE requirements, hashed token storage, and both MCP lookup paths, including legacy compatibility. Checks recorded for this change:

- `SECRET_KEY=test-secret-key-that-is-long-enough-for-oauth-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache uv run pytest tests/test_oauth_api.py tests/test_mcp_servers.py -q`
- `SECRET_KEY=test-secret-key-that-is-long-enough-for-oauth-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache ./run_tests.sh`

### DIFF
--- a/backend/app/api/mcp.py
+++ b/backend/app/api/mcp.py
@@ -55,6 +55,7 @@ from app.services.execution_cancellation import (
 )
 from app.services.global_variables_service import get_global_variables_context
 from app.services.mcp_session import mcp_session_store, mcp_sse_channels
+from app.services.oauth_tokens import oauth_token_lookup_values
 from app.services.workflow_executor import execute_workflow
 
 router = APIRouter()
@@ -144,7 +145,7 @@ async def get_mcp_user(
         now = datetime.now(timezone.utc)
         result = await db.execute(
             select(OAuthAccessToken).where(
-                OAuthAccessToken.access_token == bearer_token,
+                OAuthAccessToken.access_token.in_(oauth_token_lookup_values(bearer_token)),
                 OAuthAccessToken.revoked.is_(False),
                 OAuthAccessToken.expires_at > now,
             )
--- a/backend/app/api/mcp_servers.py
+++ b/backend/app/api/mcp_servers.py
@@ -44,6 +44,7 @@ from app.services.execution_cancellation import clear_execution as clear_active_
 from app.services.execution_cancellation import register_execution
 from app.services.global_variables_service import get_global_variables_context
 from app.services.mcp_session import mcp_session_store, mcp_sse_channels
+from app.services.oauth_tokens import oauth_token_lookup_values
 from app.services.workflow_executor import execute_workflow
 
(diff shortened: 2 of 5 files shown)


mbakgun:
Thanks a lot for putting this together @VykosMolt 🚀  This is a solid hardening pass, and the direction looks right to me. Hashing stored OAuth tokens, enforcing PKCE S256 for public clients, rejecting unsupported client auth methods, using a timing-safe PKCE comparison, and adding row locks around replay-sensitive grants all make sense.

I think this is very close. Before merging, we have two requests:

1. Could we add endpoint-level coverage for the hardened OAuth flow? In particular, a public client without PKCE should be rejected, and the authorization code/token exchange path should exercise the real /authorize and /token behavior rather than only helper-level tests.

2. Could we add coverage for the hashed bearer-token lookup paths in the MCP endpoints, including the legacy plaintext fallback? Those are the exact resource-server paths that need to keep working after the storage change.

One transition note: the legacy plaintext fallback is useful for compatibility, but existing plaintext refresh tokens can remain usable until expiry or rotation. I do not think we need to force a revoke migration in this PR unless we want the security benefit to land immediately, but it would be helpful to document the intended transition behavior in the PR description so the tradeoff is explicit.

A small optional cleanup: _validate_pkce_request validates stripped PKCE values, but the original unstripped code_challenge and code_challenge_method are stored on the auth code. Storing the normalized values would keep authorize and token exchange behavior consistent. Not a security blocker, just a consistency improvement.

One separate follow-up idea: the MCP endpoints still accept raw bearer tokens through the ?token= query param, which can leak through logs, browser history, or referrers. Since this is broader than the token-storage change and may exist for SSE compatibility, I am fine tracking that separately.

Thanks again 🙏. This is good work and I think it is close to ready. 🤗

mbakgun (approved):
Thanks for the quick follow-up. I validated the latest PR head in a clean temporary checkout.

Looks good to merge. Thanks again.

PR #155, opened 2026-06-06
Limit upload reads before buffering

VykosMolt:
## Summary

This adds a shared bounded-read helper for uploaded files and uses it in the upload paths that were reading the entire request body into memory:

- manual generated-file uploads
- vector-store document uploads
- data-table CSV imports
- voice transcription uploads

The helper reads only `file_max_size_mb + 1 byte` worth of data, returns a `413` when the configured limit is exceeded, and keeps the existing downstream storage/processing behavior for valid files.

## Verification

- `SECRET_KEY=test-secret-key-that-is-long-enough-for-upload-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache uv run pytest tests/test_upload_limits.py tests/test_file_storage.py tests/test_voice_api.py -q`
- `SECRET_KEY=test-secret-key-that-is-long-enough-for-upload-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache ./run_tests.sh`

### DIFF
--- a/backend/app/api/data_tables.py
+++ b/backend/app/api/data_tables.py
@@ -34,6 +34,7 @@ from app.models.schemas import (
     DataTableTeamShareResponse,
     DataTableUpdate,
 )
+from app.services.upload_limits import read_upload_file_limited
 
 router = APIRouter()
 
@@ -659,7 +660,7 @@ async def import_csv(
     columns = table.columns or []
     col_names = {col["name"] for col in columns}
 
-    content = await file.read()
+    content = await read_upload_file_limited(file)
     try:
         text = content.decode("utf-8-sig")
     except UnicodeDecodeError:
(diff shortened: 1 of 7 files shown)


mbakgun (approved):
Thanks for the clean hardening PR @VykosMolt  💯  This fixes the handler-level unbounded UploadFile.read() issue across Drive uploads, vector store uploads, CSV imports, and voice transcription by reading only the configured limit plus one byte and returning 413 for oversized files.

This does not fully replace request-body or temp-file limits at the ASGI/proxy layer, so we can track that separately as a follow-up. Merging this as a useful and low-risk improvement.

PR #156, opened 2026-06-06
Restrict workflow share management to owners

VykosMolt:
## Summary

This closes a workflow sharing permission gap from the audit.

User-level workflow share endpoints now match the existing team-share behavior: only the workflow owner can list, add, or remove user shares. A user who only has shared access to a workflow can still access the workflow itself, but they can no longer inspect or change the share list for someone else's workflow.

## Verification

- `SECRET_KEY=test-secret-key-that-is-long-enough-for-share-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache uv run pytest tests/test_workflow_share_permissions.py tests/test_workflow_execution_api.py tests/test_workflow_update_websocket_sync.py -q`
- `SECRET_KEY=test-secret-key-that-is-long-enough-for-share-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache ./run_tests.sh`

### DIFF
--- a/backend/app/api/workflows.py
+++ b/backend/app/api/workflows.py
@@ -1565,6 +1565,11 @@ async def list_workflow_shares(
             status_code=status.HTTP_404_NOT_FOUND,
             detail="Workflow not found",
         )
+    if workflow.owner_id != current_user.id:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Only owner can list shares",
+        )
 
     result = await db.execute(
         select(WorkflowShare, User)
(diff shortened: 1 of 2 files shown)


mbakgun (approved):
Yes, that makes sense, because there is no longer any reason for it not to be this way. 

In early days of Heym, without the team feature, this was a feature we had left open so that a user who was granted permission, other than the owner, could share with another user without having to go back to the owner every time, providing flexibility. 

Now the team feature can be used for this purpose, and this can remain 1-n; it does not need to be 1-n-n. 

Thank you 🚀

PR #158, opened 2026-06-06
Use non-mutating lint in PR checks

VykosMolt:
## Summary

This keeps PR checks read-only.

The existing frontend `lint` script still runs ESLint with `--fix` for local cleanup, but CI now uses a new `lint:check` script that runs `eslint .` without modifying files. That makes lint failures explicit in pull requests instead of letting the workflow rewrite the checkout.

## Verification

- `BUN_INSTALL_CACHE_DIR=/tmp/heym-bun-cache bun run lint:check`

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Frontend lint
         working-directory: frontend
-        run: bun run lint
+        run: bun run lint:check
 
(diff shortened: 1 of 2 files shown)


mbakgun (approved):
+1 Looks good. PR checks should be read-only and fail clearly instead of mutating the checkout; keeping `lint` for local fixes and using `lint:check` in CI is the right split.
Thanks !

PR #159, opened 2026-06-06
Make Docker log access opt in

VykosMolt:
## Summary

This makes Docker-backed log access opt-in instead of giving the backend the host Docker socket by default.

- adds `DOCKER_LOGS_ENABLED=false` as the default setting
- adds `DOCKER_LOGS_ALLOWED_EMAILS` so enabling the feature still limits access to trusted users
- makes the Docker logs endpoints return `403` before initializing a Docker client when the flag is off, the allow-list is empty, or the user is not listed
- removes the Docker socket mount from the default compose service, with an inline opt-in note
- updates the README and env examples to explain when and how to enable it

Mounting `/var/run/docker.sock` is useful for the Logs tab in some self-hosted setups, but it also gives the backend broad control over the host Docker daemon. This keeps the safer default while preserving an explicit path for users who need it and know which accounts should be allowed through.

## Verification

- `SECRET_KEY=test-secret-key-that-is-long-enough-for-log-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache uv run pytest tests/test_logs_api.py tests/test_settings.py -q`
- `SECRET_KEY=test-secret-key-that-is-long-enough-for-log-tests ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 UV_CACHE_DIR=/tmp/heym-uv-cache ./run_tests.sh`


VykosMolt:
This is the last of my first pass PRs. I would also recommend a refactor but Id await maintainer approval before I start working on that.

mbakgun:
Thank you for the PR and for tightening this default. 🚀

I think the issue is valid and important. Mounting /var/run/docker.sock gives the backend broad control over the host Docker daemon, so making Docker-backed log access opt-in is the right direction.

The access-control logic in the fix looks sound. It rejects requests before initializing the Docker client when the feature is disabled, when the allow-list is empty, or when the user email is not listed. I also validated the targeted tests in a temporary copy, and they passed. ✅

A few follow-ups would make this cleaner before merge:

1. Please document DOCKER_LOGS_ALLOWED_EMAILS in the deployment docs next to DOCKER_LOGS_ENABLED, and mention it in the Logs tab docs as a required setting when Docker logs are enabled. Suggested wording: "Set DOCKER_LOGS_ALLOWED_EMAILS to a comma-separated list of trusted user emails that are allowed to access Docker logs." 🤗
2. Consider improving the frontend error display so 403 responses show the backend detail instead of a generic status message.
3. The email allow-list assumes the listed address already belongs to a trusted account. With ALLOW_REGISTER=true and no email-ownership verification at registration, someone could self-register an unclaimed allow-listed email and gain log access. A docs note recommending that the admin account be created before enabling the feature, or that registration stays closed, would make this clearer.

Thanks for raising this up once again 🙏

VykosMolt:
Ill get right on it.

mbakgun:
Thank you for the quick follow-up and for addressing the review points.

One thing I want to revisit before merge: can we restore the Docker socket mount in the default Docker Compose service and in the docker run examples?

Removing `-v /var/run/docker.sock:/var/run/docker.sock` is a breaking change beyond the Logs tab. Docker-based MCP stdio setups can rely on that socket even when Docker logs are disabled, since the Docker CLI inside the backend still needs access to the host Docker daemon.

I think we should keep the Logs API protected by `DOCKER_LOGS_ENABLED` and `DOCKER_LOGS_ALLOWED_EMAILS`, but avoid removing the socket mount in this PR. We can document that Docker log access is gated by the new settings, while Docker-based MCP stdio tools may still use the socket mount.

If we want to make the socket mount itself opt-in later, I think that should be a separate breaking change with release notes and migration guidance.

mbakgun (approved):
Thanks for the contribution. After the merge conflict is fixed, we can merge this PR. 👑

PR #160, opened 2026-06-06
refactor(backend): workflow api

eryue0220:
## Summary

extract workflow access and execution persistence into shared services.

Move duplicated workflow access control, credential loading, sub-workflow resolution, and execution history/analytics/global-variable persistence out of `workflows.py` and trigger/API entry points into `services/workflow_access.py` and `services/execution_persistence.py`. 

No intended behavior change; add focused unit tests.

## Test plan

- [x] `SECRET_KEY=test-secret-key-for-tests-only-32-bytes ./check.sh`


mbakgun:
Thanks a lot for putting this together. The duplication you are targeting is real, especially around execution history persistence, analytics snapshots, sub-workflow histories, global-variable persistence, workflow access checks, and credential loading. Moving that logic toward shared services is a reasonable direction.

That said, this is a large cross-cutting refactor that touches many execution entry points at once. For a change with this much surface area, We(@ckakgun) would prefer that we first align on the shared service shape and the call-site contract, then land it in smaller, easier-to-review steps. That would make it much easier to confirm that each execution path keeps the same semantics.

I also noticed that the PR is not strictly behavior-preserving. In particular, some paths now persist global variables because `persist_execution_result` defaults to `persist_global_variables=True`, and the RabbitMQ path now stores `trigger_source="rabbitmq"` where it previously did not set one. These may be valid changes, but they should be explicit, intentional, and covered at the relevant call sites. This was just an example, when I reviewed submitted code, I noticed others as well. 

For now, I do not think we should merge this PR as-is, so I am going to close it. I would be happy to revisit the direction in a smaller follow-up proposal and split the implementation into narrower PRs once we agree on the design. (https://github.com/heymrun/heym/discussions)

PR #161, opened 2026-06-06
Add ASGI request body size limit

VykosMolt:
Follow-up to the upload hardening discussion in #155. That PR limits handler-level UploadFile reads; this adds an ASGI-layer request body guard so oversized bodies can be rejected before endpoint handlers buffer them.


mbakgun:
Thanks for the follow-up hardening work @VykosMolt 👑  . The request body guard itself looks sound to me 💯 : I've tested and it rejects oversized Content-Length requests before the app is called, catches chunked bodies while reading from receive, and keeps valid requests working.

One thing worth addressing: oversized requests currently bypass CORS headers because the body limit middleware short-circuits outside the CORS middleware. From the browser, a large upload from the frontend may show up as a CORS or network error instead of the clear 413 JSON response.

This does not break the security control, so I do not see it as a blocker. It would still be cleaner to either make CORS wrap this response or add the appropriate CORS headers to the 413 response path.

Also, the `test_workflow_assistant_stream.py `change looks unrelated to the body-size guard. Could you confirm whether it is covering a flaky test fix, or split it out if it is unrelated? 🙏 

And also ,  `file_max_size_mb: int = 99 (config.py) `, do you think , we should align the numbers ?

mbakgun:
Thanks !

PR #163, opened 2026-06-08
feat(frontend): add password hint

eryue0220:
## Summary

Add real-time password strength hints to the registration page. As users type, a live checklist shows each requirement (minimum length, uppercase, lowercase, and number), aligned with backend validation, so they understand what’s required before submitting.

## Test

image


ckakgun (changes requested):
Thanks for adding this @eryue0220 💯  Can we keep vertical scrolling available on mobile?

The login page still scrolls on small screens, but this change makes the register page use `h-dvh with overflow-hidden`, so on smaller phones or when the keyboard is open, the bottom of the form can become unreachable.

https://github.com/user-attachments/assets/ed5db0a2-1536-4170-95ed-64af6460b208

I think we should avoid hiding vertical overflow here. For example, keep horizontal overflow hidden only, or add overflow-y-auto so the register form remains usable on small phones.

eryue0220:
> Thanks for adding this @eryue0220 💯 Can we keep vertical scrolling available on mobile?
> 
> The login page still scrolls on small screens, but this change makes the register page use `h-dvh with overflow-hidden`, so on smaller phones or when the keyboard is open, the bottom of the form can become unreachable.
> 
>  Clipboard-20260608-052508-731.mp4 
> I think we should avoid hiding vertical overflow here. For example, keep horizontal overflow hidden only, or add overflow-y-auto so the register form remains usable on small phones.

Hi @ckakgun ,

I've update the layout for the mobile and PC, 
1. remove the limitation of height that user can scroll on small phones
2. let the main content display fully on small phones

image

image

ckakgun (approved):
thanks !

Issue #168, opened 2026-06-09
[Feature]: Add discord integration


eryue0220:
## Summary

As [discussion](https://github.com/heymrun/heym/discussions/167), this PR is to record the progress of discord integrations.

## Tracking

- [x] Discord Webhook
- [x] Discord Trigger
- [x] Discord Node for frontend
- [x] Documentation Update for Discord Webhood and Trigger
- [ ] Discord Use Case [Templates](https://heym.run/templates)

eryue0220:
Hi @mbakgun could you check is anything else that I missed in this task? And can you assign this issue to me? 

Thank you

ckakgun:
@eryue0220 Looks great, thanks for taking this on. Please go ahead and continue

ckakgun:
This development work is complete, so I am closing the issue. Thank you again. You can submit template variations anytime via https://heym.run/submit Thanks again for your efforts, and see you next time.

PR #175, opened 2026-06-10
feat(backend): add discord webhook

eryue0220:
## Summary

This pr is focus on adding discord webhook for backend 

## Related Issue

Fixed Issue: #168


ckakgun:
Hello, once the development is complete, you can remove the WIP label. The entire development can proceed through this pull request.

eryue0220:
> Hello, once the development is complete, you can remove the WIP label. The entire development can proceed through this pull request.

Hi, this pr is complete. As previously mentioned, this pr only focus on discord webhook. As for the discord trigger / discord node will split into  another PR. 

And I think it is more easily to review.

ckakgun:
As @mbakgun  mentioned yesterday in this link, https://github.com/heymrun/heym/discussions/167#discussioncomment-17224384, we actually want to handle this as a single pull request.

The reason is that there are teams/users who may clone the repository at any time and bring it up with deploy.sh, and we do not want to distribute and have people use incomplete, fragmented work.

A feature becomes mergeable for Heym once it is completed end-to-end.

We can actually continue within this pull request. The number of commits does not matter, but it is important that the work enters as a single PR.

Thank you for your understanding.

eryue0220:
> As @mbakgun mentioned yesterday in this link, [#167 (comment)](https://github.com/heymrun/heym/discussions/167#discussioncomment-17224384), we actually want to handle this as a single pull request.
> 
> The reason is that there are teams/users who may clone the repository at any time and bring it up with deploy.sh, and we do not want to distribute and have people use incomplete, fragmented work.
> 
> A feature becomes mergeable for Heym once it is completed end-to-end.
> 
> We can actually continue within this pull request. The number of commits does not matter, but it is important that the work enters as a single PR.
> 
> Thank you for your understanding.

Okay, now it's updated.

mbakgun:
Again, nice work @eryue0220 👑 Thank you so much 🤗 ! The Discord integration is wired end to end across credentials, migrations, trigger endpoint, outbound node, frontend panels, docs, and tests. Backend Discord tests pass, ruff is clean, and GitHub checks are green. 🔥

A few changes before merge:

1. Deferred interaction is acknowledged but never completed

In `backend/app/api/discord.py`, a valid interaction returns a type 5 deferred response and runs the workflow in the background, but I do not see anything sending a follow-up back to Discord using the interaction token.

For a slash command, this may mean the user keeps seeing "Bot is thinking..." and never receives a result through the interaction. The outbound `discord` node appears to post to a separate channel incoming webhook, which is not the same as replying to the interaction.

If you tested this end to end and the interaction response works correctly, please share the exact test steps/procedure and expected Discord-side result so we can verify the same behavior.

If not, please either:

a. Implement interaction follow-up support using Discord's interaction webhook endpoints, so the workflow result can be delivered back to the command, or

b. If this is intentionally a fire-and-forget trigger, adjust the behavior and docs so it does not imply request and response slash command replies. The current docs example `discordTrigger -> llm -> discord` may be read as a direct slash command reply flow. If the `discord` node is only sending through a separate incoming webhook, please clarify that in the docs.

2. Align `discord-node.md` with the actual output, preferably by returning the useful sent fields

`frontend/src/docs/content/nodes/discord-node.md` lists output fields as `status`, `message`, `username`, and `avatar_url`, but the executor currently returns only `{ status, response }`.

I think the useful sent fields should be available downstream where possible. Please either update the executor output to include `message` (the resolved message text that was sent), `username`, and `avatar_url` in addition to `status` and `response`, or update the docs to match the current output. My preference is to return the useful non-sensitive fields.

Everything else looks good to me. 🚀

eryue0220:
> mplement interaction follow-up support using Discord's interaction webhook endpoints, so the workflow result can be delivered back to the

Updated.

1. discordTrigger now sends the workflow’s final Output back to Discord as a follow-up message after the initial deferred interaction response, so interaction flows no longer stop at the deferred type: 5 ack only.
2. I aligned the regular discord node output with its documented fields by returning message, username, and avatar_url in addition to status and response. I also updated the docs to clarify the intended pattern: use discordTrigger -> ... -> output for interaction replies, and use the regular discord node only when an additional webhook post is needed.

ckakgun (approved):
Thank you so much for adding the Discord command trigger and webhook node as a message publisher.

PR #181, opened 2026-06-11
fix: centralize chat markdown rendering

Rohithmatham12:
## Summary
- Reuse the shared sanitized markdown renderer in portal chat, dashboard chat, and main chat conversation surfaces
- Remove duplicated markdown sanitizer configs that could drift across chat UIs
- Drop the unnecessary `style` attribute from the shared user-markdown allowlist
- Copy rendered plaintext, not raw markdown, from exported portal HTML assistant messages
- Escape JSON embedded in exported portal chat HTML so message text cannot break out of the inline script

## Testing
- `bun run lint:check`
- `bun run typecheck`
- `bun run build`
- `git diff --check`
- `uv run ruff check .`

### DIFF
--- a/frontend/src/components/Chat/ChatConversation.vue
+++ b/frontend/src/components/Chat/ChatConversation.vue
@@ -20,8 +20,6 @@ import {
   AudioLines,
   X,
 } from "lucide-vue-next";
-import { marked } from "marked";
-import DOMPurify from "dompurify";
 
 import type { Message, WorkflowPreview } from "@/types/chat";
 import type { CredentialListItem, LLMModel } from "@/types/credential";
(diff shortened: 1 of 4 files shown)


mbakgun:
Thanks for cleaning this up @Rohithmatham12  👑  Centralizing the markdown rendering is a good idea and should reduce sanitizer drift across the chat surfaces.

Could you also fix one remaining security issue in the portal chat HTML export path?

`plainTexts` is embedded directly inside an inline `<script>` via `JSON.stringify(plainTexts)`. If a message contains `</script><script>...</script>`, the browser HTML parser can close the script tag early and execute injected script in the exported HTML.

This looks pre-existing rather than introduced by this PR, but since this PR already touches the export plaintext behavior, it would be great to handle it here too. Please escape script-breaking sequences before embedding JSON into the inline script, for example by replacing `<`, `>`, `&`, U+2028, and U+2029, or move the data into a safer non-executable container such as a JSON script tag with properly escaped text content.

Rohithmatham12:
Thanks for catching that. I updated the export path to escape JSON before embedding it in the inline script, covering `<`, `>`, `&`, U+2028, and U+2029. I also reran lint, typecheck, build, and diff checks.

mbakgun:
Thank you 💯

PR #184, opened 2026-06-12
feat: add amazon s3 service

eryue0220:
## Summary

Adds an Amazon S3 workflow node and s3 credential type to Heym. The integration uses the standard boto3 S3 client with AWS SigV4 authentication and targets Amazon S3 only.


eryue0220:
Some of functions currently not included in this PR


Category | Examples | Status
-- | -- | --
Non-AWS storage | MinIO, Cloudflare R2, custom endpoints | Not supported
Large / multipart uploads | Multipart upload, upload_file | Not supported
Presigned URLs | generate_presigned_url | Not supported
Metadata / ACL | head_object, ACL, object tagging, lifecycle rules | Not supported
Bucket configuration | Bucket policy, CORS, versioning setup | Not supported
Versioning workflows | ListObjectVersions, delete by VersionId | Partial — version_id and delete_marker are returned, but there are no dedicated operations
Binary upload | Raw bytes / stream upload | Not supported — text upload only; download supports base64

mbakgun (changes requested):
Thanks for the PR 🤗 @eryue0220. The S3 node is a useful addition, and this looks close to merge. The backend service plus executor wiring look solid 🔥

The scope limitations you listed, such as no MinIO/R2/custom endpoints, no multipart uploads, no presigned URLs, no ACL/tagging/lifecycle ops, and text-only upload, are reasonable v1 choices rather than bugs, as long as they stay documented.

A few concrete items I think should be addressed before merge:

1. Editing an existing S3 credential can fail. On edit, the dialog blanks `s3AccessKeyId` and `s3SecretAccessKey` but defaults `s3Region` to `us-east-1`. Since `s3Region` is non-empty, it is included in the `hasConfigChange` check, so the frontend can treat the config as changed and send an update with an empty access key and secret. The backend then rejects it during validation. The result is that even renaming an S3 credential can fail unless the user re-enters the key and secret. Suggested fix: populate `s3Region` from the stored credential on edit, or exclude the default region from change detection.

2. Docs slug mismatch. The added file is `amazon-s3-node.md`, but the manifest and node help link use the slug `s3-node`. The docs loader resolves a slug to `{slug}.md`, so `/docs/nodes/s3-node` will not find the page. Renaming the file to `s3-node.md` should fix it.

3. The AI Workflow Builder prompt does not appear to be updated for the new S3 node. The assistant learns node schemas, operations, required fields, credential rules, and examples from `workflow_dsl_prompt.py`, but this PR does not add S3 guidance there. That means a request like “add an AWS S3 create bucket node” may not be generated reliably. Please add the S3 node schema to the workflow DSL prompt, including supported `s3Operation` values such as `createBucket`, required fields like `credentialId` and `s3Bucket`, and at least one example workflow.

A few minor points, not blockers:

- The region dropdown is a fixed list and appears to be missing newer AWS regions, including `ap-southeast-6` and `ap-east-2`, which are listed in the current AWS Regions documentation: https://docs.aws.amazon.com/global-infrastructure/latest/regions/aws-regions.html. Since the field is a locked select, users in those regions cannot pick them.
- `get_object` reads the full body into memory, so very large objects can be costly, especially with base64 output. Worth a note or a size guard later.
- For `putObject`, the expression field navigation counts 4 fields, but the 4th step reopens the body field instead of Content Type, since Content Type has no expression ref.

eryue0220:
> Thanks for the PR 🤗 @eryue0220. The S3 node is a useful addition, and this looks close to merge. The backend service plus executor wiring look solid 🔥
> 
> The scope limitations you listed, such as no MinIO/R2/custom endpoints, no multipart uploads, no presigned URLs, no ACL/tagging/lifecycle ops, and text-only upload, are reasonable v1 choices rather than bugs, as long as they stay documented.
> 
> A few concrete items I think should be addressed before merge:
> 
> 1. Editing an existing S3 credential can fail. On edit, the dialog blanks `s3AccessKeyId` and `s3SecretAccessKey` but defaults `s3Region` to `us-east-1`. Since `s3Region` is non-empty, it is included in the `hasConfigChange` check, so the frontend can treat the config as changed and send an update with an empty access key and secret. The backend then rejects it during validation. The result is that even renaming an S3 credential can fail unless the user re-enters the key and secret. Suggested fix: populate `s3Region` from the stored credential on edit, or exclude the default region from change detection.
> 2. Docs slug mismatch. The added file is `amazon-s3-node.md`, but the manifest and node help link use the slug `s3-node`. The docs loader resolves a slug to `{slug}.md`, so `/docs/nodes/s3-node` will not find the page. Renaming the file to `s3-node.md` should fix it.
> 3. The AI Workflow Builder prompt does not appear to be updated for the new S3 node. The assistant learns node schemas, operations, required fields, credential rules, and examples from `workflow_dsl_prompt.py`, but this PR does not add S3 guidance there. That means a request like “add an AWS S3 create bucket node” may not be generated reliably. Please add the S3 node schema to the workflow DSL prompt, including supported `s3Operation` values such as `createBucket`, required fields like `credentialId` and `s3Bucket`, and at least one example workflow.
> 
> A few minor points, not blockers:
> 
> * The region dropdown is a fixed list and appears to be missing newer AWS regions, including `ap-southeast-6` and `ap-east-2`, which are listed in the current AWS Regions documentation: https://docs.aws.amazon.com/global-infrastructure/latest/regions/aws-regions.html. Since the field is a locked select, users in those regions cannot pick them.
> * `get_object` reads the full body into memory, so very large objects can be costly, especially with base64 output. Worth a note or a size guard later.
> * For `putObject`, the expression field navigation counts 4 fields, but the 4th step reopens the body field instead of Content Type, since Content Type has no expression ref.

Thanks for your review, it's been updated.

mbakgun:
> > Thanks for the PR 🤗 @eryue0220. The S3 node is a useful addition, and this looks close to merge. The backend service plus executor wiring look solid 🔥
> > The scope limitations you listed, such as no MinIO/R2/custom endpoints, no multipart uploads, no presigned URLs, no ACL/tagging/lifecycle ops, and text-only upload, are reasonable v1 choices rather than bugs, as long as they stay documented.
> > A few concrete items I think should be addressed before merge:
> > 
> > 1. Editing an existing S3 credential can fail. On edit, the dialog blanks `s3AccessKeyId` and `s3SecretAccessKey` but defaults `s3Region` to `us-east-1`. Since `s3Region` is non-empty, it is included in the `hasConfigChange` check, so the frontend can treat the config as changed and send an update with an empty access key and secret. The backend then rejects it during validation. The result is that even renaming an S3 credential can fail unless the user re-enters the key and secret. Suggested fix: populate `s3Region` from the stored credential on edit, or exclude the default region from change detection.
> > 2. Docs slug mismatch. The added file is `amazon-s3-node.md`, but the manifest and node help link use the slug `s3-node`. The docs loader resolves a slug to `{slug}.md`, so `/docs/nodes/s3-node` will not find the page. Renaming the file to `s3-node.md` should fix it.
> > 3. The AI Workflow Builder prompt does not appear to be updated for the new S3 node. The assistant learns node schemas, operations, required fields, credential rules, and examples from `workflow_dsl_prompt.py`, but this PR does not add S3 guidance there. That means a request like “add an AWS S3 create bucket node” may not be generated reliably. Please add the S3 node schema to the workflow DSL prompt, including supported `s3Operation` values such as `createBucket`, required fields like `credentialId` and `s3Bucket`, and at least one example workflow.
> > 
> > A few minor points, not blockers:
> > 
> > * The region dropdown is a fixed list and appears to be missing newer AWS regions, including `ap-southeast-6` and `ap-east-2`, which are listed in the current AWS Regions documentation: [docs.aws.amazon.com/global-infrastructure/latest/regions/aws-regions.html](https://docs.aws.amazon.com/global-infrastructure/latest/regions/aws-regions.html). Since the field is a locked select, users in those regions cannot pick them.
> > * `get_object` reads the full body into memory, so very large objects can be costly, especially with base64 output. Worth a note or a size guard later.
> > * For `putObject`, the expression field navigation counts 4 fields, but the 4th step reopens the body field instead of Content Type, since Content Type has no expression ref.
> 
> Thanks for your review, it's been updated.


Thanks for the update. I rechecked the latest commit and most of the requested items look fixed now.

The S3 credential edit flow looks good, the AI Workflow Builder prompt now includes S3 guidance and a create bucket example, and the missing AWS regions were added. 

One small docs link issue still remains before approval: the docs file, manifest, and cross-links now use `amazon-s3-node`, but `nodeDocSlugMap` in `PropertiesPanel.vue` still maps S3 to `s3-node`. That means the S3 node help link from the canvas will still navigate to `/docs/nodes/s3-node`, which no longer exists.

Please update:

```ts
s3: "s3-node",
```

to:

```ts
s3: "amazon-s3-node",
```

After that, this should be good from my side.

eryue0220:
> `/docs/nodes/s3-node`,

Ohhh, missing this one.  Updated now.

mbakgun (approved):
I'm incredibly grateful! This is such an amazing contribution! 🔥

PR #187, opened 2026-06-12
fix: require credential access when cloning vector stores

Rohithmatham12:
## Summary
- Require the current user to still have access to a vector store's backing Qdrant credential before cloning the store
- Store the access-checked credential id on the cloned vector store instead of blindly copying the source store's credential id
- Add regression coverage for inaccessible backing credentials and for successful clones using the access-checked credential

## Testing
- `uv run ruff check .`
- `uv run ruff format --check .`
- `git diff --check`
- `SECRET_KEY=test-secret-key-for-tests-only-32-bytes ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000001 uv run python -m unittest tests.test_vector_store_credential_access` (2 passed)
- `SECRET_KEY=test-secret-key-for-tests-only-32-bytes ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000001 uv run python -m unittest tests.test_workflow_execution_api.WorkflowExecutorDataTableNodeTests.test_shared_vector_store_uses_backing_credential_without_separate_share` (1 passed)
- `SECRET_KEY=test-secret-key-for-tests-only-32-bytes ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000001 ./run_tests.sh` (1007 passed)

### DIFF
--- a/backend/app/api/vector_stores.py
+++ b/backend/app/api/vector_stores.py
@@ -105,6 +105,14 @@ def get_vector_store_service_from_config(config: dict):
     )
 
 
+async def _get_store_credential_config(
+    store: VectorStore,
+    user_id: uuid.UUID,
+    db: AsyncSession,
+) -> tuple[Credential, dict]:
+    return await get_credential_config(store.credential_id, user_id, db)
+
+
 @router.get("", response_model=list[VectorStoreListResponse])
 async def list_vector_stores(
     current_user: User = Depends(get_current_user),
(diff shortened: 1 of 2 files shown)


mbakgun (approved):
Thanks for the fix @Rohithmatham12 🎉 I reviewed the current behavior and this looks like a valid issue.

Before this change, a user with access to a shared vector store could clone it even if they did not have access to the backing Qdrant credential. The cloned store would then become owned by that user while still referencing the original private credential. That is an unintended privilege extension, and it matters in multi-user or team setups.

The fix looks correct: clone now goes through the same credential access check used elsewhere, so it no longer blindly carries over a credential the user cannot access.

One small note: this fixes future clones, but it does not address any existing cloned stores that may already reference inaccessible credentials. That is fine for this PR. 🚀

PR #198, opened 2026-06-14
fix: log exceptions swallowed by drain_bg_futures (#197)

### LINKED ISSUE #197: bug: drain_bg_futures` silently swallows exceptions from background sub-workflow execution
## Steps to Reproduce
1. Create a workflow that contains an **Execute Sub-Workflow (Do Not Wait)** node.
2. Configure the sub-workflow so it will fail at runtime (e.g. reference a non-existent node output, use a credential that returns an error, or introduce a deliberate exception in a Code node).
3. Execute the parent workflow.
4. Observe the parent workflow completes without any error indication.

## Expected Behaviour
When a background sub-workflow raises an exception, the error should be logged so that operators can diagnose failures. The parent workflow can still complete successfully (since the sub-workflow was fire-and-forget), but the failure must be visible — at minimum as a `WARNING` log line referencing the exception.

## Actual Behaviour
`drain_bg_futures` in `backend/app/services/workflow_executor.py` (around line 2486) contains a bare `except Exception: pass`:

```python
try:
    fut.result()
    if done_event is not None:
        done_event.wait()
except Exception:
    pass  # exception silently discarded
```

Any exception raised by a background sub-workflow — including crashes, DB errors, and cancellation failures from `done_event.wait()` — is completely discarded with no log output, no metric, and no trace entry. The only observable symptom is missing side effects in the sub-workflow's output.

Yes. The fix is a two-line change: name the exception and emit a `logger.warning`. A draft PR is ready.

syedalijaseem:
## Summary
  - `drain_bg_futures` had a bare `except Exception: pass` that silently discarded all exceptions from
  background sub-workflow execution
  - Named the exception and added a `logger.warning` so failures are visible in logs
  - Parent workflow behaviour is unchanged — fire-and-forget semantics preserved

  ## Changes
  `backend/app/services/workflow_executor.py` — 2 lines changed

  ```diff
  -            except Exception:
  -                pass
  +            except Exception as exc:
  +                logger.warning("drain_bg_futures: background sub-workflow raised: %s", exc)

  Test plan
  - [x] uv run ruff format --check . — clean
  - [x] uv run ruff check . — clean
  - [x] Full backend test suite (1156 tests) — all pass

  Closes #197

### DIFF
--- a/backend/app/services/workflow_executor.py
+++ b/backend/app/services/workflow_executor.py
@@ -2486,8 +2486,8 @@ class WorkflowExecutor:
                 fut.result()
                 if done_event is not None:
                     done_event.wait()
-            except Exception:
-                pass
+            except Exception as exc:
+                logger.warning("drain_bg_futures: background sub-workflow raised: %s", exc)
 
     @staticmethod
     def _record_bg_sub_workflow_done(
--- a/backend/app/services/workflow_executor.py
+++ b/backend/app/services/workflow_executor.py
@@ -2484,10 +2484,15 @@ class WorkflowExecutor:
                 done_event = None
             try:
                 fut.result()
+            except Exception as exc:
+                logger.warning(
+                    "drain_bg_futures: background sub-workflow raised: %s",
+                    exc,
+                    exc_info=True,
+                )
+            finally:
                 if done_event is not None:
                     done_event.wait()
-            except Exception as exc:
-                logger.warning("drain_bg_futures: background sub-workflow raised: %s", exc)
 
     @staticmethod
     def _record_bg_sub_workflow_done(
--- a/backend/tests/test_drain_bg_futures.py
+++ b/backend/tests/test_drain_bg_futures.py
@@ -1,0 +1,87 @@
+"""Regression tests for drain_bg_futures exception handling.
+
+Verifies that when a background sub-workflow future raises:
+  1. A warning is logged (with exc_info).
+  2. done_event.wait() still executes (via finally), so the done-callback
+     has finished recording status="error" before drain_bg_futures returns.
+"""
+
+from __future__ import annotations
+
+import threading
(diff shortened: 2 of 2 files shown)


mbakgun:
Thanks for the PR @syedalijaseem ⭐️ The issue is valid as an observability gap, and the direction is right. A couple of things make me think it does not fully close #197 yet.

`drain_bg_futures()` now logs after `fut.result()` raises, but on that exception path it still skips `done_event.wait()`. That wait is what guarantees the done callback has finished, and the done callback is what records the background sub-workflow as `status="error"` in `sub_workflow_executions`.

Since `result()` can observe the future as finished before callbacks have completed, the streaming path that serializes right after `drain_bg_futures()` can still race and miss the error entry. So the log line is an improvement, but the underlying "failure not reliably visible" problem can still happen on the path described in the issue.

Suggested tweaks:

- Move `done_event.wait()` into a `finally` block so it always waits, whether `fut.result()` raises or not.
- Log with `exc_info=True` so operators get the traceback, not just `str(exc)`.
- Add a small `assertLogs` regression test around a failing background future, asserting both the warning and that the `status="error"` sub-workflow execution is recorded before serialization.

Thanks again for the fix. With those small changes, I think this would fully cover the issue.

syedalijaseem:
Updated per your feedback:

1. Moved `done_event.wait()` into a `finally` block so it always runs whether `fut.result()` raises or not
2. Added `exc_info=True` to `logger.warning` for full traceback
3. Added `backend/tests/test_drain_bg_futures.py` with two regression tests:
   - asserts the WARNING is logged when a background future fails
   - asserts `status="error"` is recorded in `sub_workflow_executions` before drain returns

mbakgun (approved):
Thank you so much ⭐️ 🎉

PR #200, opened 2026-06-15
fix: block undo/redo while workflow execution is in progress (#199)

### LINKED ISSUE #199: bug: Undo/redo overwrites canvas node and edge state during live workflow execution
## Steps to Reproduce
1. Open a workflow with several nodes in the editor.
2. Make one or more edits so the undo history is non-empty (`historyIndex > 0`).
3. Start executing the workflow (click Run).
4. While the execution is streaming results, press **Ctrl+Z** (undo) or **Ctrl+Y** (redo).
5. Observe the canvas.

## Expected Behaviour
Undo and redo should be blocked while a workflow execution is in progress. The canvas state should remain stable until execution completes.

## Actual Behaviour
`undo()` and `redo()` in `frontend/src/stores/workflow.ts` (lines 290–312) each guard only against an empty history:

```typescript
// undo()
if (!canUndo.value) return;  // no isExecuting check

// redo()
if (!canRedo.value) return;  // no isExecuting check
```

Both functions immediately overwrite `nodes.value` and `edges.value` via `JSON.parse(JSON.stringify(...))` regardless of whether `isExecuting` is `true`. Since the running executor reads these reactive refs, replacing them mid-execution corrupts the canvas state — node positions, connections, and execution overlays become inconsistent for the remainder of the run.

`isExecuting` is already defined in the same store (`ref(false)`, line 62) and toggled by `startExecution`/`stopExecution` — it is simply not consulted by the two undo/redo functions.

Yes. The fix is a two-line change — add `|| isExecuting.value` to the guard in each function. A draft PR is ready.

syedalijaseem:
## Summary

- `undo()` and `redo()` in `frontend/src/stores/workflow.ts` only checked `canUndo`/`canRedo` — they did not consult `isExecuting`, so pressing Ctrl+Z or Ctrl+Y while a workflow was running would immediately overwrite the reactive `nodes.value` and `edges.value` refs with a prior history snapshot.
- Since the canvas executor reads those same reactive refs, replacing them mid-execution corrupts the running graph: node positions, edge connections, and execution-state overlays become inconsistent for the remainder of the run.
- Fix is a two-line guard change. `isExecuting` was already defined in the same store (`ref(false)`, line 62) and is correctly set/unset by `startExecution`/`stopExecution` — `undo()`/`redo()` simply never referenced it.

## Changes

- `frontend/src/stores/workflow.ts`
  - `undo()` (line 291): `if (!canUndo.value) return;` → `if (!canUndo.value || isExecuting.value) return;`
  - `redo()` (line 303): `if (!canRedo.value) return;` → `if (!canRedo.value || isExecuting.value) return;`

## Test Plan

- [ ] Open a workflow and make several edits so the undo history is non-empty.
- [ ] Start executing the workflow (click Run / trigger execution).
- [ ] While execution is in flight (execution indicator active), press **Ctrl+Z** — canvas nodes and edges must not change.
- [ ] While execution is in flight, press **Ctrl+Y** — canvas nodes and edges must not change.
- [ ] Let execution complete, then press **Ctrl+Z** — undo should work as normal.
- [ ] Press **Ctrl+Y** after undoing — redo should work as normal.
- [ ] Run frontend lint and typecheck: `bun run lint && bun run typecheck` — both must pass.

Closes #199

### DIFF
--- a/frontend/src/stores/workflow.ts
+++ b/frontend/src/stores/workflow.ts
@@ -288,7 +288,7 @@ export const useWorkflowStore = defineStore("workflow", () => {
   }
 
   function undo(): void {
-    if (!canUndo.value) return;
+    if (!canUndo.value || isExecuting.value) return;
 
     isUndoRedo.value = true;
     historyIndex.value--;
@@ -300,7 +300,7 @@ export const useWorkflowStore = defineStore("workflow", () => {
   }
 
   function redo(): void {
-    if (!canRedo.value) return;
+    if (!canRedo.value || isExecuting.value) return;
 
     isUndoRedo.value = true;
     historyIndex.value++;


ckakgun (approved):
Thank you 💯

PR #213, opened 2026-06-17
feat: add supabase service

eryue0220:
## Summary

Adds an supabase workflow node and supabase credential type to Heym.


ckakgun:
Thanks for the solid implementation @eryue0220 🚀 Supabase seems really handy addition to the Heym ecosystem so we're glad for you effort on this ⭐  The Supabase node is well covered overall, the service and executor tests are strong, and the DSL plus in-app docs updates are much appreciated.

Before merge, could you please address the items below?

1. Fix logical `or` / `and` filter encoding

`SupabaseService._compile_filter_params()` strips the outer parentheses from logical filters, so `{"or":[...]}` is sent as:

```text
or=status.eq.active,and(status.eq.pending,score.gte.5)
```

PostgREST requires the value to keep the outer parentheses, for example:

```text
or=(status.eq.active,and(status.eq.pending,score.gte.5))
```

Please keep the outer parens and update `test_select_rows_supports_logical_groups` to assert the PostgREST-compatible output. Right now the test asserts the broken form, so it passes while the feature can fail against a real Supabase project.

2. Strip whitespace from `supabase_url`

`SupabaseService.__init__` trims only trailing slashes, not whitespace. A pasted URL with a trailing space will fail confusingly. The API key already calls `.strip()`; please do the same for the URL:

```python
self._base_url = str(self._config.get("supabase_url", "")).strip().rstrip("/")
```

3. Add a guard for `supabaseLimit = "0"`

Fetching all rows with no upper bound can exhaust memory in the executor worker on large tables. Please add a hard cap or another bounded strategy, and raise a readable `ValueError` when the fetch would exceed it.

4. Use strict equality in `PropertiesPanel.vue`

The sequence guards in `loadSupabaseTablesForSelectedNode` and `loadSupabaseColumnsForSelectedNode` use `!=`, while the rest of the codebase uses `!==`. Please switch these to `!==` for consistency.

5. Add a Supabase expression evaluator test

AGENTS.md calls out that changes to expression-evaluated fields should extend `backend/tests/test_expression_evaluator_service.py`. `supabaseTable`, `supabaseFilter`, `supabaseSchema`, and `supabaseSelectColumns` all go through expression resolution, so please add at least one case verifying a `$ref` expression resolves correctly inside a Supabase node config.

6. Revisit `CredentialResponse` shape

Adding `supabase_url` and `supabase_schema` directly to the shared response means every credential type now returns Supabase-specific null fields. Please move integration-specific public edit fields under a generic `public_fields` object so future credentials do not keep expanding the shared model.

7. Either remove or justify `grid-layout-plus.d.ts`

`frontend/src/types/grid-layout-plus.d.ts` looks unrelated to Supabase. If it is only a typecheck fix, please note that in the PR description; otherwise it would be cleaner in a separate PR.

8. Align docs on supported filters

`reference/integrations.md` says Supabase currently supports exact-match filters, but `supabase-node.md` and the DSL prompt document operator and logical filters. Please update the integrations wording to match the implemented behavior.

eryue0220:
> Thanks for the solid implementation @eryue0220 🚀 Supabase seems really handy addition to the Heym ecosystem so we're glad for you effort on this ⭐ The Supabase node is well covered overall, the service and executor tests are strong, and the DSL plus in-app docs updates are much appreciated.
> 
> Before merge, could you please address the items below?
> 
> 1. Fix logical `or` / `and` filter encoding
> 
> `SupabaseService._compile_filter_params()` strips the outer parentheses from logical filters, so `{"or":[...]}` is sent as:
> 
> ```
> or=status.eq.active,and(status.eq.pending,score.gte.5)
> ```
> 
> PostgREST requires the value to keep the outer parentheses, for example:
> 
> ```
> or=(status.eq.active,and(status.eq.pending,score.gte.5))
> ```
> 
> Please keep the outer parens and update `test_select_rows_supports_logical_groups` to assert the PostgREST-compatible output. Right now the test asserts the broken form, so it passes while the feature can fail against a real Supabase project.
> 
> 2. Strip whitespace from `supabase_url`
> 
> `SupabaseService.__init__` trims only trailing slashes, not whitespace. A pasted URL with a trailing space will fail confusingly. The API key already calls `.strip()`; please do the same for the URL:
> 
> ```python
> self._base_url = str(self._config.get("supabase_url", "")).strip().rstrip("/")
> ```
> 
> 3. Add a guard for `supabaseLimit = "0"`
> 
> Fetching all rows with no upper bound can exhaust memory in the executor worker on large tables. Please add a hard cap or another bounded strategy, and raise a readable `ValueError` when the fetch would exceed it.
> 
> 4. Use strict equality in `PropertiesPanel.vue`
> 
> The sequence guards in `loadSupabaseTablesForSelectedNode` and `loadSupabaseColumnsForSelectedNode` use `!=`, while the rest of the codebase uses `!==`. Please switch these to `!==` for consistency.
> 
> 5. Add a Supabase expression evaluator test
> 
> AGENTS.md calls out that changes to expression-evaluated fields should extend `backend/tests/test_expression_evaluator_service.py`. `supabaseTable`, `supabaseFilter`, `supabaseSchema`, and `supabaseSelectColumns` all go through expression resolution, so please add at least one case verifying a `$ref` expression resolves correctly inside a Supabase node config.
> 
> 6. Revisit `CredentialResponse` shape
> 
> Adding `supabase_url` and `supabase_schema` directly to the shared response means every credential type now returns Supabase-specific null fields. Please move integration-specific public edit fields under a generic `public_fields` object so future credentials do not keep expanding the shared model.
> 
> 7. Either remove or justify `grid-layout-plus.d.ts`
> 
> `frontend/src/types/grid-layout-plus.d.ts` looks unrelated to Supabase. If it is only a typecheck fix, please note that in the PR description; otherwise it would be cleaner in a separate PR.
> 
> 8. Align docs on supported filters
> 
> `reference/integrations.md` says Supabase currently supports exact-match filters, but `supabase-node.md` and the DSL prompt document operator and logical filters. Please update the integrations wording to match the implemented behavior.

Hi @ckakgun , thanks for your review and I've updated.

mbakgun (approved):
thx so much @eryue0220 once again ⭐️

PR #218, opened 2026-06-18
fix: skip resume_hitl_request_in_background for non-pending or expire…

### LINKED ISSUE #217: bug: resume_hitl_request_in_background` proceeds without checking if HITL request has expired
## Steps to Reproduce

1. Create a workflow with a HITL (Human-in-the-Loop) node.
2. Trigger the workflow so a `HITLRequest` is created with `status="pending"` and an `expires_at` timestamp.
3. Let the request expire naturally (do not act on it before `expires_at`).
4. At some point after expiry — e.g. due to a delayed background task from a prior server restart or a race on a busy host — `resume_hitl_request_in_background` is called with the expired request's `id`.
5. The function proceeds to execute `resume_workflow_execution` and writes `history_entry.status = "success"` to the database.

## Expected Behaviour

`resume_hitl_request_in_background` should abort early when the `HITLRequest` has `status="expired"` or `expires_at < now`. An expired approval should never resume workflow execution.

## Actual Behaviour

`resume_hitl_request_in_background` (`backend/app/services/hitl_service.py`, lines 336–418) fetches the `HITLRequest` and immediately proceeds past three early-return guards (`hitl_request is None`, `workflow is None`, `credentials_owner_id missing`) without ever checking the request's status or expiry:

```python
async def resume_hitl_request_in_background(request_id: uuid.UUID) -> None:
    async with async_session_maker() as db:
        result = await db.execute(select(HITLRequest).where(HITLRequest.id == request_id))
        hitl_request = result.scalar_one_or_none()
        if hitl_request is None:          # only guard present
            return
        # ← no check for hitl_request.status or hitl_request.expires_at
        ...
        resumed_result = await asyncio.to_thread(resume_workflow_execution, ...)
```

The expiry guard exists in the synchronous viewing path (`ensure_hitl_request_is_viewable`, line 52) and the action path (`ensure_hitl_request_is_actionable`, line 421), but is absent from the background resume path. A stale background task — queued before server restart, delayed under load, or triggered by a retry — can therefore execute an approval that was legitimately expired, producing incorrect workflow state.

Yes. The fix is a single guard added after the `hitl_request is None` check:

```python
if hitl_request.status != "pending" or hitl_request.expires_at < datetime.now(timezone.utc):
    return
```

This mirrors the logic already used by `ensure_hitl_request_is_actionable`. A PR with a regression test is ready.

syedalijaseem:
## Summary

- `resume_hitl_request_in_background` in `backend/app/services/hitl_service.py` fetched a `HITLRequest` from the database and proceeded directly to resume workflow execution without checking whether the request was still `pending`.
- A stale background task — queued before a server restart, delayed under load, or retried — could call this function after the request had already expired, been accepted, or been refused, causing the workflow to resume from an approval that was no longer valid.
- All other HITL code paths (`ensure_hitl_request_is_viewable`, `ensure_hitl_request_is_actionable`, `get_hitl_request_by_token`) check status and/or expiry. The background resume path was the only one that skipped this guard.
- Fix is a single two-line early return added immediately after the `hitl_request is None` check.

## Changes

- `backend/app/services/hitl_service.py`
  - After line 341 (`if hitl_request is None: return`): added
    ```python
    if hitl_request.status != "pending" or hitl_request.expires_at < datetime.now(timezone.utc):
        return
    ```
- `backend/tests/test_hitl_resume_expiry_guard.py` (new file)
  - Regression tests asserting that `resume_hitl_request_in_background` does not call `resume_workflow_execution` when the request has `status="expired"`, `"accepted"`, or `"refused"`, and when `status="pending"` but `expires_at` is in the past.

## Test Plan

- [ ] Run the full backend test suite: `SECRET_KEY=test-secret-key-for-tests-only-32-bytes ./run_tests.sh` — all 1291 tests must pass.
- [ ] Manually trigger a HITL workflow and let the request expire before acting on it. Confirm that a delayed resume call (or server restart + retry) does not incorrectly resume the workflow.
- [ ] Manually trigger a HITL workflow, approve it normally. Confirm that `resume_hitl_request_in_background` completes successfully for the `pending` → accepted transition.
- [ ] Run backend linting: `uv run ruff check . && uv run ruff format --check .` — must pass clean.

Closes #217

### DIFF
--- a/backend/app/services/hitl_service.py
+++ b/backend/app/services/hitl_service.py
@@ -339,7 +339,7 @@ async def resume_hitl_request_in_background(request_id: uuid.UUID) -> None:
         hitl_request = result.scalar_one_or_none()
         if hitl_request is None:
             return
-        if hitl_request.status != "pending" or hitl_request.expires_at < datetime.now(timezone.utc):
+        if hitl_request.status != "resolved" or not hitl_request.decision:
             return
 
         workflow = await db.get(Workflow, hitl_request.workflow_id)
--- a/backend/app/services/hitl_service.py
+++ b/backend/app/services/hitl_service.py
@@ -339,6 +339,8 @@ async def resume_hitl_request_in_background(request_id: uuid.UUID) -> None:
         hitl_request = result.scalar_one_or_none()
         if hitl_request is None:
             return
+        if hitl_request.status != "pending" or hitl_request.expires_at < datetime.now(timezone.utc):
+            return
 
         workflow = await db.get(Workflow, hitl_request.workflow_id)
         history_entry = await db.get(ExecutionHistory, hitl_request.execution_history_id)
--- a/backend/tests/test_hitl_resume_expiry_guard.py
+++ b/backend/tests/test_hitl_resume_expiry_guard.py
@@ -1,54 +1,56 @@
-"""Regression test: resume_hitl_request_in_background must not resume non-pending or expired requests."""
+"""Regression test: resume_hitl_request_in_background must only resume resolved requests with a decision."""
 
 import uuid
-from datetime import datetime, timedelta, timezone
 from unittest import IsolatedAsyncioTestCase
 from unittest.mock import AsyncMock, MagicMock, patch
 
 
-def _mock_session(hitl_request: MagicMock) -> MagicMock:
+def _make_mock_db(hitl_request: MagicMock) -> AsyncMock:
     mock_result = MagicMock()
     mock_result.scalar_one_or_none.return_value = hitl_request
     mock_db = AsyncMock()
     mock_db.execute.return_value = mock_result
+    mock_db.get.return_value = None  # workflow not found → safe early exit after guard
+    return mock_db
+
+
+def _make_session(mock_db: AsyncMock) -> MagicMock:
     mock_session = AsyncMock()
     mock_session.__aenter__.return_value = mock_db
     mock_session.__aexit__.return_value = False
     return mock_session
 
 
+def _make_request(status: str, decision: str | None) -> MagicMock:
+    req = MagicMock()
+    req.status = status
+    req.decision = decision
+    return req
+
+
 class TestResumeHitlExpiryGuard(IsolatedAsyncioTestCase):
-    async def test_non_pending_status_is_not_resumed(self) -> None:
-        for status in ("expired", "accepted", "refused"):
-            hitl_request = MagicMock()
-            hitl_request.status = status
-            hitl_request.expires_at = datetime.now(timezone.utc) + timedelta(hours=1)
-
-            with (
-                patch(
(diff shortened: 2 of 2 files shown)


ckakgun:
Thanks for the fix and the tests @syedalijaseem 🚀 

I agree that `resume_hitl_request_in_background` is missing a defensive guard, so the underlying issue is valid. However, the guard in this PR breaks the normal HITL flow. 💯 

In the actual flow, both callers (`submit_hitl_decision` in `api/hitl.py` and the AI assistant path) set `HITLRequest.status = "resolved"` and commit before scheduling `resume_hitl_request_in_background`. The background task then reloads the request, where the status is already `"resolved"`. Since the proposed check returns whenever the status is not `"pending"`, a normally approved, edited, or refused request would never resume the workflow. I confirmed this locally: with the patch applied, a resolved request with a future expiry exits before reaching `resume_workflow_execution`.

A related note: `"accepted"` and `"refused"` are not real `HITLRequest.status` values in this codebase. They are decision/output concepts stored via the `decision` field, so the regression tests assert states that never actually occur. The real lifecycle is closer to `pending -> resolved`, or `expired`.

A safer guard should allow only a resolved request with a valid decision to resume, while rejecting unresolved pending requests and explicitly expired requests, for example:

```python
if hitl_request.status != "resolved" or not hitl_request.decision:
    return
```

The tests should also include the `status="resolved"` happy path, since that is exactly the path this PR currently regresses. 🙏

syedalijaseem:
Fixed per your feedback. The guard now checks `status != "resolved"` or missing `decision`, instead of the old pending + expiry logic.

I also rewrote the tests. The old ones were testing `"accepted"` and `"refused"` as statuses (which don't exist on `HITLRequest`) and were patching `resume_workflow_execution` as the signal, which required threading through the full mock chain. The new tests use `db.get` as the sentinel: if the guard fires, `db.get` is never reached; if the guard passes, it is.

Four cases are covered: resolved + decision (passes), pending (blocked), expired (blocked), and resolved without a decision (blocked).

mbakgun:
Thanks for the quick update and for reworking the tests. 🤗

I rechecked the latest commit and this now looks correct. The guard now matches the actual HITL lifecycle by allowing only a resolved request with a decision to resume, while blocking pending, expired, or unresolved requests. I also confirmed that the normal resolved happy path is no longer blocked.

One small non-blocking cleanup request before we merge: could you update the test happy path to use `decision="accept"` instead of `decision="approve"`? The guard only checks truthiness, so the current test still passes, but `accept` matches the real decision values used by the app (`accept`, `edit`, `refuse`) and keeps the test closer to production behavior.

After that, this looks good to merge. 🚀

mbakgun (approved):
thank you @syedalijaseem ! 🚀

PR #220, opened 2026-06-18
fix: snapshot label_to_output under lock in _build_context to prevent…

### LINKED ISSUE #219: bug: `_build_context` iterates shared dicts without a lock under concurrent node execution
## Steps to Reproduce

1. Create a workflow with multiple nodes that can execute in parallel (e.g. two independent LLM nodes feeding into a merge node).
2. Run the workflow so that `WorkflowExecutor` uses its `ThreadPoolExecutor` to execute nodes concurrently.
3. While one thread is inside `_build_context` iterating `self.label_to_output`, another thread calls `store_node_output` and mutates the same dict.
4. Observe a `RuntimeError: dictionary changed size during iteration` crash in the executor thread.

## Expected Behaviour

`_build_context` should read a consistent snapshot of `label_to_output` and `_wrapped_label_output_cache`. Concurrent writes from other node threads should not affect an in-progress context build.

## Actual Behaviour

The window is narrow under CPython's GIL and requires near-simultaneous node completions, so this surfaces as an intermittent hang rather than a consistent crash. However, the race is real and the fix is two lines.

`_build_context` (`backend/app/services/workflow_executor.py`) iterates directly over `self.label_to_output` and checks `self._wrapped_label_output_cache` without holding `self.lock`:

```python
# BEFORE — unsafe reads
for label, data in self.label_to_output.items():   # ← mutated by concurrent store_node_output
    if label in self._wrapped_label_output_cache:   # ← mutated by concurrent store_node_output
        ...
```

`store_node_output` is called by every completing node thread and mutates both dicts under `self.lock`. Because `_build_context` acquires no lock before iterating, a concurrent write from another thread can resize the dict mid-iteration and raise an unhandled `RuntimeError`, silently killing the executor thread and causing the workflow to hang or produce incomplete output.

Yes. The fix takes a locked snapshot of both dicts at the top of `_build_context` before any iteration begins:

```python
with self.lock:
    label_to_output_snapshot = dict(self.label_to_output)
    wrapped_cache_snapshot = dict(self._wrapped_label_output_cache)
```

All subsequent reads in both the `current_node_id` and `else` branches use the snapshots instead of the live dicts. A regression test that hammers `_build_context` from the main thread while a writer thread calls `store_node_output` in a tight loop (500 iterations) is included.

syedalijaseem:
## Summary

- `_build_context` in `WorkflowExecutor` iterated directly over `self.label_to_output` and read `self._wrapped_label_output_cache` without holding `self.lock`.
- Under concurrent node execution, `store_node_output` (called from worker threads) mutates both dicts under the lock. A race between a writer thread and an in-progress `_build_context` call could resize the dict mid-iteration, raising `RuntimeError: dictionary changed size during iteration` and silently killing the executor thread.
- Fix: take a single locked snapshot of both dicts at the top of `_build_context` and read only the snapshots throughout. The snapshot itself is O(n) in the number of node outputs and held only for the duration of the `dict()` copy — the lock is not held during any heavy computation or wrapping.

## Changes

- `backend/app/services/workflow_executor.py`
  - In `_build_context`: added `with self.lock:` block to snapshot `self.label_to_output` and `self._wrapped_label_output_cache` before iteration.
  - Both the `current_node_id` branch and the `else` branch now read from `label_to_output_snapshot` / `wrapped_cache_snapshot` instead of the live dicts.

- `backend/tests/test_build_context_concurrency.py` (new file)
  - Regression test: starts a writer thread calling `store_node_output` in a tight loop while the main thread calls `_build_context({})` 500 times.
  - Would reliably raise `RuntimeError` before the fix; passes cleanly after.

## Test Plan

- [x] Regression test `test_no_dict_mutation_error_under_concurrent_store` added and passes.
- [x] Full backend test suite: 1291 passed, 0 failures (`SECRET_KEY=test-secret-key-for-tests-only-32-bytes uv run pytest tests/ -x -q`).
- [x] Ruff format and lint: `2 files left unchanged, All checks passed`.

Closes #219


mbakgun (approved):
thx  ! <3

PR #228, opened 2026-06-18
feat: add gitHub service and node

eryue0220:
This pr is to add new github credential service and e2e node


mbakgun:
@eryue0220 ⭐️ Thanks for this PR 🤗 , it is a really nice addition and fits Heym well 🔥. I went through it and the overall structure, tests, and checks are in good shape.

There is one thing I would like to sort out before we merge. The `dispatchWorkflowAndWait` operation reads `workflow_run_id` from the dispatch response, but by default GitHub's workflow dispatch endpoint returns `204 No Content` with no body. It only returns `200` with the run id when the request payload includes `return_run_details: true`.

Because of that, this operation can fail against the real GitHub API even though the current test passes because it mocks a `200` response with a run id.

Could you handle this in one of these ways?

1. Add `return_run_details: true` to the dispatch payload when waiting is required, and update the test so it also covers the real default `204` case.
2. Or dispatch first, then locate the new run by polling the workflow runs list filtered by branch, event, and time.

A few smaller things would also be good to clean up:

- `getUserIssues`: the "mentioned" filter likely needs `filter=mentioned` rather than passing `mentioned` as a separate query param while `filter` stays `assigned`, so the UI filter probably will not behave as expected right now.
- The GitHub Enterprise `base_url` docs do not match the backend. The code keeps the existing base URL on edit, but the docs say it gets removed. Worth aligning one with the other.
- Frontend validation for `updateRelease` requires a tag name, but the backend allows updating just the body, name, or draft fields. The current validation may block valid update flows.
- The new github node isn't registered in the DSL prompt (**backend/app/services/workflow_dsl_prompt.py**). Comparable nodes like supabase, slack, telegram, bigquery, and s3 all have "type": "..." entries there, so the AI workflow generator can emit them. Without this, prompt-based workflow generation won't know the github node exists. Could you add an entry with its operations and key fields?

Everything else looks solid 🚀. Once the wait behavior is fixed, this should be safe to merge. Thanks again for putting this together.

eryue0220:
> @eryue0220 ⭐️ Thanks for this PR 🤗 , it is a really nice addition and fits Heym well 🔥. I went through it and the overall structure, tests, and checks are in good shape.
> 
> There is one thing I would like to sort out before we merge. The `dispatchWorkflowAndWait` operation reads `workflow_run_id` from the dispatch response, but by default GitHub's workflow dispatch endpoint returns `204 No Content` with no body. It only returns `200` with the run id when the request payload includes `return_run_details: true`.
> 
> Because of that, this operation can fail against the real GitHub API even though the current test passes because it mocks a `200` response with a run id.
> 
> Could you handle this in one of these ways?
> 
> 1. Add `return_run_details: true` to the dispatch payload when waiting is required, and update the test so it also covers the real default `204` case.
> 2. Or dispatch first, then locate the new run by polling the workflow runs list filtered by branch, event, and time.
> 
> A few smaller things would also be good to clean up:
> 
> * `getUserIssues`: the "mentioned" filter likely needs `filter=mentioned` rather than passing `mentioned` as a separate query param while `filter` stays `assigned`, so the UI filter probably will not behave as expected right now.
> * The GitHub Enterprise `base_url` docs do not match the backend. The code keeps the existing base URL on edit, but the docs say it gets removed. Worth aligning one with the other.
> * Frontend validation for `updateRelease` requires a tag name, but the backend allows updating just the body, name, or draft fields. The current validation may block valid update flows.
> * The new github node isn't registered in the DSL prompt (**backend/app/services/workflow_dsl_prompt.py**). Comparable nodes like supabase, slack, telegram, bigquery, and s3 all have "type": "..." entries there, so the AI workflow generator can emit them. Without this, prompt-based workflow generation won't know the github node exists. Could you add an entry with its operations and key fields?
> 
> Everything else looks solid 🚀. Once the wait behavior is fixed, this should be safe to merge. Thanks again for putting this together.

Thanks for the review, I've updated.

ckakgun:
Thanks for the updates @eryue0220. The dispatch-and-wait fix via polling the workflow runs list looks good, and the new DSL prompt entry is a nice addition.

There are two things left before this is good to merge.

1. GitHub Enterprise `base_url` docs still contradict the backend

A few docs still say that saving a rotated token without re-entering `base_url` removes the Enterprise endpoint. The backend does not behave that way anymore.

`merge_credential_config_for_update` starts from the existing config and only overwrites `base_url` when a non-empty value is sent. The frontend also omits `base_url` from the payload when the field is left empty. So the existing Enterprise base URL is preserved on edit.

Could you update the remaining docs to say that the existing `base_url` is preserved and does not need to be re-entered when rotating the token?

I still see this stale wording in:

- `frontend/src/docs/content/nodes/github-node.md`
- `frontend/src/docs/content/reference/credentials.md`
- `frontend/src/docs/content/reference/integrations.md`
- `frontend/src/docs/content/tabs/credentials-tab.md`

2. Double-clicking a GitHub node does not open the expression dialog

For other nodes, double-clicking on the canvas opens the properties panel and auto-opens the expression expand dialog on the node's primary field, so users can write expressions with autocomplete right away.

For the GitHub node, double-click currently opens the panel but does not open any expression dialog. The GitHub fields already use `ExpressionInput` refs, so the missing piece is wiring the node into the same double-click flow as the other node types.

The two places to update are:

- `frontend/src/components/Canvas/WorkflowCanvas.vue`: `handleNodeDoubleClick` maps each `nodeType` to a `fieldToFocus`, but there is no `github` case.
- `frontend/src/components/Panels/PropertiesPanel.vue`: `openPrimaryExpandDialogForSelectedNode` has per-node branches that call `someInputRef.value?.openExpandDialog()`, but there is no `github` branch.

Could you add a `github` branch in both places and choose the primary field based on `githubOperation`, similar to the `rag` branch?

Suggested mapping:

- `createComment` -> `githubCommentBodyExpressionInputRef`
- `createIssue` / `updateIssue` -> `githubBodyExpressionInputRef`
- `upsertFile` -> `githubFileContentExpressionInputRef`
- `createRelease` / `updateRelease` -> `githubBodyExpressionInputRef`
- `dispatchWorkflow` / `dispatchWorkflowAndWait` -> `githubWorkflowInputsExpressionInputRef`
- fallback -> `githubOwnerExpressionInputRef`

Please keep the same `tryOpenDialog` retry pattern used by the other branches. Also, note that owner is hidden for `getUserIssues` and `inviteUser`, so the fallback should try a couple of alternatives, such as `githubOrganizationExpressionInputRef`, `githubMentionedExpressionInputRef`, `githubRepoExpressionInputRef`, or `githubWorkflowIdExpressionInputRef`, so those operations do not no-op.

One small implementation note: `handleNodeDoubleClick` currently types `event.node.data` with only `nodeType` and `ragOperation`, so it will also need to include `githubOperation` or use a safe cast when selecting the GitHub focus field.

Once these two are fixed, this should be ready to merge. Thanks again.

eryue0220:
> 2\. Double-clicking a GitHub node does not open the expressi

Thanks, I've updated.

ckakgun (approved):
Thanks. To avoid unnecessary back and forth, we will merge this PR. The expression dialog is not working exactly as intended, so we will fix it in the next version

Thanks again for the node 👑

PR #233, opened 2026-06-19
feat: add fe e2e test

eryue0220:
This pr is to add e2e test for frontend. But it's not 100% coverage for the whole codebase.


mbakgun:
Thank you so much for this, really great work 🙏 ⭐️ 🎉 @eryue0220 . This is exactly the kind of feature and quality step we want to have. Frontend has had no test coverage at all, and a real browser plus real backend smoke and regression suite over the key paths is genuinely valuable. I also appreciate how clear you were about the scope and the limits in the docs.

A few notes:

The three fixes are all solid. The Vite `/register` proxy bypass in particular closes a real dev bug, since `/register` is both an SPA route and an OAuth registration endpoint. The 401 interceptor change and the pricing sync flag, with its backend test, are both correct and low risk.

One request before merge: please do not add E2E to `./check.sh` by default. We want to keep `./check.sh` as the fast local validation path for lint, typecheck, formatting, and backend tests. The E2E suite should still run in GitHub Actions, and locally it can stay available through `./run_e2e.sh`.

Also, please update `AGENTS.md`. Right now it still says "Frontend: No tests yet (add when needed)" and the Testing section mainly describes backend pytest. Since this PR establishes frontend E2E as part of the workflow, `AGENTS.md` should reflect that. Please add:

- Frontend E2E tests use Playwright and live in `frontend/e2e/`.
- Run them locally with `./run_e2e.sh`.
- GitHub Actions runs the E2E suite for PRs.
- Meaningful frontend workflow or UX changes should add or extend E2E coverage when the behavior is testable.

Once `check.sh` keeps the fast local path and `AGENTS.md` is in sync, this is good to go. Thanks again. 👑 🚀

mbakgun:
Thanks, the `check.sh` revert and the `AGENTS.md` update look great.

One small leftover: `frontend/src/docs/content/getting-started/running-and-deployment.md` still says that `./check.sh` includes the E2E suite and mentions `SKIP_E2E`. That no longer applies now that E2E runs separately through `./run_e2e.sh`.

Could you update that paragraph so the docs and `AGENTS.md` stay in sync? 🚀

eryue0220:
> Thanks, the `check.sh` revert and the `AGENTS.md` update look great.
> 
> One small leftover: `frontend/src/docs/content/getting-started/running-and-deployment.md` still says that `./check.sh` includes the E2E suite and mentions `SKIP_E2E`. That no longer applies now that E2E runs separately through `./run_e2e.sh`.
> 
> Could you update that paragraph so the docs and `AGENTS.md` stay in sync? 🚀

Thank you for your check the details, I've updated.

PR #272, opened 2026-07-01
fix: make plugin path safety checks cross-platform

VeraPyuyi:
## Summary
- replace Windows-fragile plugin path prefix checks with `Path.relative_to`
- keep plugin zip-slip and icon traversal protection intact
- add regression coverage for nested plugin archive members

## Root cause
The previous containment checks compared resolved paths as strings with a hard-coded `/` separator. On Windows, resolved paths use `\`, so valid plugin members could be reported as unsafe.

## Validation
- `python -m pytest tests/test_plugin_store.py tests/test_plugin_loader.py tests/test_plugin_manifest.py tests/test_plugin_dsl_prompt.py tests/test_plugin_node.py -q`
- `python -m ruff check app/services/plugin_store.py app/api/plugins.py tests/test_plugin_store.py`
- `python -m ruff format --check app/services/plugin_store.py app/api/plugins.py tests/test_plugin_store.py`
- `git diff --check`
- `bash ./check.sh` via Git Bash: 1729 tests, 1729 passed

### DIFF
--- a/backend/app/api/plugins.py
+++ b/backend/app/api/plugins.py
@@ -53,7 +53,9 @@ def _safe_icon_path(plugin_id: str, filename: str) -> Path | None:
         return None
     package_root = (plugin_store.plugins_root() / plugin_id).resolve()
     candidate = (package_root / filename).resolve()
-    if candidate != package_root and not str(candidate).startswith(str(package_root) + "/"):
+    try:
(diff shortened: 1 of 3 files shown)


ckakgun (approved):
Yes, thank you for your contribution. 🎊

PR #274, opened 2026-07-01
feat(frontend): add searchable select input

eryue0220:
# Summary

Convert node operation dropdowns to the shared searchable select UI across the remaining E2E-covered node property panels. Which can help user find the operation, especially when the operation list is too long. 

UI is in following the pictures:

0272d8622643e521ae842ffe03132b35

### DIFF
--- a/frontend/e2e/canvas-workflow-executions.spec.ts
+++ b/frontend/e2e/canvas-workflow-executions.spec.ts
@@ -439,7 +439,7 @@ test("shows an error workflow when an HTTP API call fails", async ({ page }) => 
     [
       workflowNode("http_error", "http", 120, 160, {
         label: "fetchBrokenApi",
-        curl: "curl -X GET http://",
+        curl: "curl -X GET http://127.0.0.1:9/e2e-http-error",
       }),
       workflowNode("output_unreachable", "output", 380, 160, {
         label: "unreachableOutput",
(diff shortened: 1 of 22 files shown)


mbakgun:
Thanks for the PR @eryue0220  🙏  The searchable operation select is a genuinely useful UX improvement, especially for nodes with long operation lists like GitHub, Notion, Linear, and S3 👑

Before merging, could you address two small things and one related UX ask? 🤗

**Duplicate operation lists.** The new grouped lists duplicate the old flat operation lists for Linear, GitHub, ClickHouse, Notion, and S3. The old `*OperationOptions` are still returned from the controller but no longer used by the node components. Could we either remove the now-unused flat lists or derive one from the other, so there is a single source of truth? That will avoid drift when new operations get added later.

**Unrelated E2E change.** In `canvas-workflow-executions.spec.ts`, the curl was changed from `http://127.0.0.1:9/e2e-http-error` to `http://`. This looks unrelated to the searchable select work, and it also shifts the test from a failed HTTP call to an invalid URL scenario, which weakens what the test covers. Could you revert it, or explain why it is needed?

**Execute node target workflow selector.** The Execute node already loads available target workflows into its selector, but it still uses the old plain `Select`. Could we also move that target workflow selector to the new searchable select pattern? That would make selecting a target workflow easier when the workspace has many workflows. 

Everything else looks good on my side. Typecheck, lint, and build all passed on my local merge snapshot. Thanks again for the improvement! 🚀

eryue0220:
> canvas-workflow-executions.spec.ts

Okay, I've updated.

ckakgun:
Thank you so much 🙏

PR #277, opened 2026-07-01
fix: clarify plugin node panel states

VeraPyuyi:
## Summary
- show distinct Plugin node panel states for loading, disabled plugin subsystem, plugin API errors, no installed plugins, missing package binding, missing/disabled packages, and missing plugin node keys
- preserve existing generated field rendering when a valid plugin package and plugin node key are available
- add Playwright coverage with mocked `/api/plugins` responses for the new panel states and a valid plugin fixture

## Validation
- `cd frontend && bun run lint:check`
- `cd frontend && bun run typecheck`
- `./check.sh` (`1729 tests, 1729 passed`)

## E2E note
- Added `frontend/e2e/plugin-node-properties.spec.ts`.
- Targeted local run `./run_e2e.sh e2e/plugin-node-properties.spec.ts` was blocked because Docker daemon is not running and Docker Desktop is not installed/found on this machine.

## Review
- Initial review found missing `pluginNodeKey` was still falling back to the first plugin node; fixed with an explicit missing-key state and E2E case.
- Final review reported no Critical, Important, or Minor findings.

### DIFF
--- a/frontend/e2e/plugin-node-properties.spec.ts
+++ b/frontend/e2e/plugin-node-properties.spec.ts
@@ -213,22 +213,20 @@ test("shows when a plugin node is not bound to a package", async ({ page }) => {
   }
 });
 
-test("shows when a plugin node key is missing", async ({ page }) => {
+test("falls back to the first action node without a plugin node key", async ({ page }) => {
   const workflowId = await openWithPlugins(page, {
     pluginId: "acme-crm",
   }, [pluginFixture]);
 
   try {
-    await expect(page.getByTestId("plugin-node-status")).toHaveText(
-      'Plugin "Acme CRM" is missing a plugin node key.',
-    );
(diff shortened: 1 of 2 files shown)


mbakgun (commented):
Thanks for this one, and for the earlier plugin PRs too. Nice follow-up work and good test coverage. 🎉🤗

Two things to fix before merge:

1. The frontend and backend now disagree on a missing `pluginNodeKey`. This PR shows a "missing plugin node key" state and hides the config fields, but the backend `resolve_node_function` still falls back to the first node of the matching kind and runs it. So such a node is unconfigurable in the panel while still executing. Please align the two sides (either both fall back, or both treat it as an error).

2. `nodeDef`'s `?? nodes[0]` fallback is now dead code since `statusMessage` blanks the fields. Please remove it.

Once those are in, good to merge. 🚀

ckakgun:
thank you ! :) 🚀

PR #292, opened 2026-07-07
fix: thread-safe node input reads and JSON parse error handling

isheng-eqi:
## Summary

Fix two categories of runtime safety issues in the workflow execution engine.

## Problem 1: Race condition in node input reads (severity: high)

`get_node_inputs_for_edges` and `get_node_inputs` read shared dicts (`node_outputs`, `node_execution_contexts`, `skipped_nodes`) without acquiring `self.lock`, while `store_node_output` writes these dicts under lock and `_build_context` correctly snapshots under lock. During parallel node execution, this creates a classic write-under-lock / read-without-lock race:
- `store_node_output` (line 2751): `with self.lock: self.node_outputs[node_id] = output`
- `get_node_inputs_for_edges` (line 2373): `source_id in self.node_outputs` — no lock
- `_build_context` (line 5476): `with self.lock: ...snapshot...` — correct pattern

**Fix:** Snapshot `node_outputs`, `node_execution_contexts`, and `skipped_nodes` under `self.lock` before iteration, matching the existing `_build_context` pattern.

## Problem 2: Missing JSON parse error handling (severity: medium)

`llm_node.py:179` and `agent_node.py:46` call `self._parse_json_output()` without try-except in non-batch JSON output mode. If the LLM returns empty text or malformed JSON, the exception propagates through the retry loop without preserving the LLM trace context (`trace_id`). The batch mode at `llm_node.py:163-168` already handles this correctly per-item.

**Fix:** Wrap `_parse_json_output` in try-except. When `trace_id` is available, raise `NodeTraceableExecutionError` (already imported in both files) to preserve the trace context. Fall back to `ValueError` with clear message when no trace_id.

## Changes

- `backend/app/services/workflow_executor.py`: Lock-protect reads in `get_node_inputs_for_edges` and `get_node_inputs`
- `backend/app/services/node_execution/nodes/llm_node.py`: Add try-except for non-batch JSON parse
- `backend/app/services/node_execution/nodes/agent_node.py`: Add try-except for non-batch JSON parse

### DIFF
--- a/backend/app/services/node_execution/nodes/agent_node.py
+++ b/backend/app/services/node_execution/nodes/agent_node.py
@@ -43,11 +43,18 @@ def execute(ctx: NodeExecutionContext) -> object:
     pending_meta = output.pop("_hitl_pending", None)
     if agent_json_output_enabled and not output.get("error"):
         agent_output = output
-        parsed = self._parse_json_output(str(output.get("text", "")))
-        if isinstance(parsed, dict):
-            output = dict(parsed)
-        else:
-            output = {"value": parsed}
+        try:
+            parsed = self._parse_json_output(str(output.get("text", "")))
+            if isinstance(parsed, dict):
+                output = dict(parsed)
+            else:
+                output = {"value": parsed}
+        except Exception as exc:
+            if trace_id:
+                raise NodeTraceableExecutionError(
+                    f"Agent JSON parse error: {exc}", trace_id
+                ) from exc
+            raise ValueError(f"Agent JSON parse error: {exc}") from exc
         if agent_output.get("fallbackUsed") is not None:
             output["fallbackUsed"] = agent_output["fallbackUsed"]
         if agent_output.get("model"):
--- a/backend/app/services/node_execution/nodes/condition_node.py
+++ b/backend/app/services/node_execution/nodes/condition_node.py
@@ -25,20 +25,22 @@ def execute(ctx: NodeExecutionContext) -> object:
 
         if result:
             loop_back_targets = self._loop_back_targets_for_source_handle(node_id, "true")
-            self.skip_branch_targets_preserving_shared_downstream(
-                node_id,
-                active_targets=true_targets,
-                inactive_targets=false_targets,
-                active_exclude_node_ids=loop_back_targets,
-                inactive_stop_node_ids=loop_back_targets,
-            )
+            with self.lock:
+                self.skip_branch_targets_preserving_shared_downstream(
+                    node_id,
+                    active_targets=true_targets,
+                    inactive_targets=false_targets,
+                    active_exclude_node_ids=loop_back_targets,
+                    inactive_stop_node_ids=loop_back_targets,
+                )
         else:
             loop_back_targets = self._loop_back_targets_for_source_handle(node_id, "false")
-            self.skip_branch_targets_preserving_shared_downstream(
-                node_id,
-                active_targets=false_targets,
-                inactive_targets=true_targets,
-                active_exclude_node_ids=loop_back_targets,
-                inactive_stop_node_ids=loop_back_targets,
-            )
+            with self.lock:
+                self.skip_branch_targets_preserving_shared_downstream(
+                    node_id,
+                    active_targets=false_targets,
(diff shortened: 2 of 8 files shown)


ckakgun:
Thanks for the fix @isheng-eqi 🙏  This looks like a valid bug fix and the implementation is well targeted. 🚀 

Before merge, please add committed backend tests. Heym / AGENTS.md requires tests for behavior changes, and this PR currently adds none.

Requested coverage:
- LLM and Agent JSON parse failures preserve trace_id when available, and raise a clear ValueError without one.
- Two parallel upstream nodes feeding one downstream node collect both inputs.
- Loop execution and re-execution still work with the input snapshot change.
- output allowDownstream returns early while downstream continues.
- Background / do-not-wait execution still records downstream results.

One lock-discipline concern: `skipped_nodes` is now read via a locked snapshot, but some writes still happen outside `self.lock`, for example branch routing updates. Please either lock those writes too, or document/test why those writes cannot overlap the reads.

Otherwise this is a clean fix. Thanks again. 🌸

isheng-eqi:
Thanks for the thorough review @ckakgun! All feedback addressed:

**Lock discipline fix**: Added `with self.lock` protection around all `skipped_nodes` mutations that can overlap with the locked snapshot reads:
- Branch routing in `execute_node_parallel` (both success and error paths)
- Condition/switch `skip_branch_targets_preserving_shared_downstream` calls
- `reset_nodes_for_execution` and `prepare_loop_for_reexecution` discard operations

**Tests**: Added 9 tests covering all 5 requested scenarios in `test_workflow_executor_thread_safety.py`:
1. LLM & Agent JSON parse failures preserve trace_id → `NodeTraceableExecutionError`; raise `ValueError` without trace_id
2. Two parallel upstream nodes (branchA/branchB) feed one downstream merge node — verifies both inputs collected correctly
3. Loop with 3-iteration array — verifies re-execution works with the input snapshot change, all 4 loop results (3 iterations + 1 done) recorded
4. Output node with `allowDownstream: true` — verifies it completes and workflow returns success
5. Wait node with `doNotWait: true` — verifies downstream nodes still record results

All 20 tests pass (9 new + 11 existing branching regression). Ruff lint clean.

Let me know if anything else is needed!

ckakgun:
Thanks for the update 💯 I pulled the latest head (b554625) and reran the checks 🙏

Runtime behavior looks good: 1932 backend unit tests pass, 81 Playwright e2e tests pass, `ruff check .` passes, and the local smokes for JSON trace failures plus repeated two-parallel-upstream joins pass 👍

Two things before merge:

1. `uv run ruff format --check .` still fails because `backend/tests/test_workflow_executor_thread_safety.py` would be reformatted. Please run Ruff format and push the formatting-only diff, since this blocks the check gate.

2. The `skipped_nodes` lock fix still seems to cover only part of the write paths. The locked condition/switch blocks in `workflow_executor.py` are under `if self.test_mode and node_data.get("pinnedData") is not None`, so they cover the pinned/test-mode preview path. Normal execution dispatches through the modular handlers, and those handlers still call `skip_branch_targets_preserving_shared_downstream` or mutate `skipped_nodes` without `self.lock`, for example condition, loop, switch, and disable-node handlers.

Tests pass, so this does not look like an immediate functional break under CPython, but the claim that `skipped_nodes` writes are lock-protected does not hold for the actual runtime handler paths. Please lock those handler writes too, or document why they cannot overlap with the locked snapshot reads.

isheng-eqi:
Good catches, both fixed:

1. **Ruff format**: reformatted — `ruff format --check` now passes on all changed files.

2. **Handler lock coverage**: good catch on the test_mode path. Extended `with self.lock` to the actual modular handlers that write `skipped_nodes` during normal execution:
   - `condition_node.py` — both true/false branch paths
   - `switch_node.py` — selected-handle branch path
   - `loop_node.py` — both loop/done branch paths
   - `disable_node_node.py` — `skipped_nodes.add()`

All 20 tests still pass, ruff format and lint clean. Let me know if anything else!

ckakgun (approved):
Thanks, the latest commit addresses both points. Looks merge-safe from my side. 🤗

PR #350, opened 2026-07-17
feat:  harden JSON Schema data contracts for node outputs

eryue0220:
# Summary

- Strengthen data-contract validation with Draft 2020-12 JSON Schema checks, safety limits (size/depth/nodes/patterns), format checkers, and safe `$ref` resolution.
- Reject invalid schemas at save/create time across workflows, templates, dashboards, and AI-generated workflows (422), and validate provider-specific structured-output schemas for LLM/agent nodes.
- Share an `OutputContractSection` in the properties panel with live frontend schema validation; strip internal/runtime metadata before contract checks during execution.


check.sh
```diff
@@ -7,6 +7,7 @@ REPO_ROOT="$(cd "$(dirname "$0")" && pwd)"
 # Shared with the CI "check" job (see .github/workflows/pr-checks.yml) so the
 # line-limit guard stays identical in both places.
 sh "$REPO_ROOT/scripts/check-file-line-limits.sh"
+sh "$REPO_ROOT/scripts/sync-json-schema-vocabulary.sh" --check
```
eryue0220, line 10: keep a single source of truth at `shared/json-schema-vocabulary.json` to decrease a potential bug in case of missing updating the file in `backend` or `frontend`

ckakgun:
Thank you @eryue0220  for the substantial amount of work and care you put into this PR. The implementation includes thoughtful validation, testing, and security hardening. 🙏 

After reviewing it, we have decided not to merge this PR and will close it. This decision is about product direction and backward compatibility, not the quality of your work.

The main reasons are:

> - This is a broad new output contract feature rather than a focused bug fix.
> - It changes batch execution from per-item partial success to failing the entire node. This affects existing users even when they have not configured an output contract, and may discard otherwise successful results.
> - The stricter validation is applied retroactively. Existing workflows using currently accepted schemas could become impossible to save, restore, or execute without a migration or compatibility path.
> - The same strict schema rules are applied across providers, although providers do not necessarily support the same JSON Schema subset.
> - The feature is exposed across many nodes, but the related DSL guidance, documentation, and UI state handling are not complete enough for a behavioral change of this scope.
> - To clarify one aspect, the $ref support here is limited to definitions within the same JSON Schema. It does not provide access to previous node results.

We do not want to introduce customer-visible behavior changes of this scope without first agreeing on the intended product behavior, backward compatibility strategy, provider-specific rules, and migration plan. For that reason, we are closing the PR rather than requesting further changes to the current implementation.

Thank you again for the contribution and for the significant effort you put into it. ⭐

eryue0220:
> Thank you @eryue0220 for the substantial amount of work and care you put into this PR. The implementation includes thoughtful validation, testing, and security hardening. 🙏
> 
> After reviewing it, we have decided not to merge this PR and will close it. This decision is about product direction and backward compatibility, not the quality of your work.
> 
> The main reasons are:
> 
> > * This is a broad new output contract feature rather than a focused bug fix.
> > * It changes batch execution from per-item partial success to failing the entire node. This affects existing users even when they have not configured an output contract, and may discard otherwise successful results.
> > * The stricter validation is applied retroactively. Existing workflows using currently accepted schemas could become impossible to save, restore, or execute without a migration or compatibility path.
> > * The same strict schema rules are applied across providers, although providers do not necessarily support the same JSON Schema subset.
> > * The feature is exposed across many nodes, but the related DSL guidance, documentation, and UI state handling are not complete enough for a behavioral change of this scope.
> > * To clarify one aspect, the $ref support here is limited to definitions within the same JSON Schema. It does not provide access to previous node results.
> 
> We do not want to introduce customer-visible behavior changes of this scope without first agreeing on the intended product behavior, backward compatibility strategy, provider-specific rules, and migration plan. For that reason, we are closing the PR rather than requesting further changes to the current implementation.
> 
> Thank you again for the contribution and for the significant effort you put into it. ⭐

Got it, thanks for your sharing.

PR #401, opened 2026-07-22
Apply OpenCode changes

heym-coder:
## Change Summary

OpenCode did not return a change summary. Files changed in this pull request:

- `backend/app/api/workflows.py`
- `backend/app/models/schemas.py`
- `frontend/src/services/api.ts`
- `frontend/src/views/DashboardView.vue`

### DIFF
--- a/backend/app/api/workflows.py
+++ b/backend/app/api/workflows.py
@@ -38,6 +38,7 @@ from app.db.models import (
 from app.db.session import async_session_maker, get_db
 from app.models.schemas import (
     ActiveExecutionItem,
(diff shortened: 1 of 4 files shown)


mbakgun:
Closing in favor of #403, which fixes the two root causes behind this PR: the coding agent quit right before capturing the screenshot (now finished with an extra pass so the screenshot is captured before the PR is created), and "update the open PR" opened a new PR because publishMode was open_pr and the LLM-generated branch never matched. #403 adds a robust update_existing_pr and a new open_or_update_pr mode.

PR #408, opened 2026-07-24
fix: playwright typecheck error

eryue0220:
# Summary 

1. Move the `DATABASE_URL` check out of top-level `playwright.config.ts` into the backend webServer command, so config load / typecheck no longer throws when the env var is unset.
2. Include `playwright.config.ts` in frontend TypeScript checks and add Node types in `tsconfig.node.json` so the Playwright config is covered by bun run typecheck.


mbakgun (approved):
Thank you 🙏

PR #409, opened 2026-07-24
feat: Harden Agent tool observability with lifecycle records, safe payloads, and OTel spans

eryue0220:
# Summary

1. Normalize Agent `tool_calls` with stable `tool_call_id`, lifecycle status (`success` / `error` / `pending` / `timeout` / `cancelled`), timestamps, and aggregate `tool_metrics`, without changing execution results or breaking historical traces.
2. Sanitize persisted tool request/response payloads (redaction + size/depth limits); keep live execution and HITL resume state untouched.
3. Emit opt-in OpenTelemetry child spans `heym.agent.tool.execute` under the Agent node span, with tool identity/status attributes and no raw args/results by default.
4. Surface pending/timeout/cancelled and tool metrics in Traces/Chat UI, and update docs/tests accordingly.


mbakgun:
Thanks for the thoughtful observability hardening @eryue0220  👑  The lifecycle model, redaction direction, and documentation are a good foundation, and CI is green on my side as well. A few things need to be addressed before merge 💯 

1. Please preserve bounded `_generated_files` metadata during sanitization. A large skill result can consume the payload budget before that key is reached, after which `agent_node.py` cannot collect the generated file information. This silently breaks the documented `$agentLabel._generated_files[0].id` flow and generated-file downloads.

Repro:

```
entry = {"name": "run_skill", "result": {
    "output": "A" * 40000,
    "logs": ["L" * 2000 for _ in range(30)],
    "_generated_files": [{"id": "abc-123", "filename": "r.pdf", "download_url": "..."}],
}}
sanitize_persisted_tool_entry(entry)
# result keys: ['output', 'logs', '_truncated']
```

Please reserve the minimal file metadata regardless of preceding payload size and add a regression test.

2. Please reconcile the pending HITL entry on resume. The pending record is saved into `agent_state.tool_calls`, copied back as `initial_tool_calls` in `workflow_executor.py` around `resume_tool_calls` (currently around line 5050), and then carried unchanged into the completed result and `tool_metrics`. A run that was approved and finished successfully should not still report a pending tool call.

3. Please keep the total trace safety limit, but allocate it fairly across tool records. The current single shared budget can be consumed by early calls, after which later calls collapse into `[PAYLOAD_TRUNCATED]` and disappear from the debugging surface. A per-entry allocation, while retaining a bounded total cap, would preserve each call's identity, status, and useful bounded payload.

4. Please either wire the Chat lifecycle statuses end to end, or drop the Chat UI claim. `ChatToolCall.vue` and `ToolCallRecord` support pending, timeout, and cancelled, but the chat SSE producer still emits only success and error, so those branches are unreachable in real chat traffic.

5. Please finish the observability wiring and coverage. `backend/app/api/config.py` in `get_observability_status` still advertises only workflow and node spans, despite the new Agent tool span. The tracing test creates the Agent tool span in isolation rather than proving it is a child of the Agent node span. Since this adds a third tracing seam in `llm_service.py`, the repository tracing guidance should also be updated or the placement should be explicitly justified.

A few smaller follow-ups:

- A cancellation arriving just after a successful tool result replaces the actual result with a cancelled payload. That loses useful audit and debugging evidence for work that completed.
- Legacy entries with `result.error` but no explicit status are currently counted as successes in `tool_metrics`.
- `tool_metrics` is computed twice: once in `_attach_tool_metrics` and again in `WorkflowExecutor._execute_agent_node`. The second value overwrites the first and also adds a zero-count metrics block to otherwise tool-less agent results. This should have a single owner.
- The final `on_tool_call(entry)` event was removed from `_commit_completed_tool_slots`. This fixes the duplicate-row issue because the old phase-less entry had no `tool_call_id` and could not merge in `workflow.ts:1543`, but it also means the live Debug panel now receives only a summary instead of the tool result. If this is an intentional privacy boundary, please document the behavior; otherwise, consider streaming a bounded sanitized result.
- Adding focused Playwright coverage for the new Chat and Traces behavior would satisfy the repository's practical UI-test expectation.

One note, not a request: `traceSteps.test.ts` is not currently run by check.sh or CI, but that is a pre-existing repository-wide gap rather than something this PR should own.

Happy to re-review once the first five items are addressed.  🙏

mbakgun (changes requested):
Happy to re-review once the items are addressed.

eryue0220:
> Happy to re-review once the items are addressed.

Thanks for the review, I've update the previous comments.

mbakgun:
Thanks for the follow-up. The generated-file preservation, HITL reconciliation, fair trace allocation, metrics cleanup, Debug payload, and OTel wiring now look good.



I found two remaining issues before I can approve.



1. Please restore cancellation checks after each completed tool.



`should_abort()` now runs only while a HITL pause is pending, so a cancellation that arrives after a normal tool completes is not observed until the next outer iteration. If the model returned several tool calls in one turn, later HTTP, MCP, or skill tools can still run after the user pressed Stop.



The fix should preserve the completed tool entry and its successful result, then stop the loop before starting the next tool. Please also make sure `_build_error_result` does not rewrite that completed entry as cancelled. `classify_tool_failure_status("Workflow execution cancelled", explicit_status="success")` still resolves to `cancelled`, so simply restoring the per-tool check would otherwise reintroduce the result-loss bug.



Please add a regression test that verifies both that the completed result remains successful and that a later tool in the same turn is not invoked.



2. Please derive Chat lifecycle status from the structured tool result or an explicit status, rather than from the display summary.



At the current head, 20 production call sites invoke `_tool_end_yield`, and only two pass an explicit status. The summary heuristic is therefore the normal path, not an exception. For example:



```text

"Created task 'Fix request timeout' on board 'Infra'" -> timeout

"Card 'Cancelled deployment': active, 2 comment(s), 1 run(s)" -> cancelled

"Card 'Foo': error, 0 comment(s), 0 run(s)" -> success

```



These are presentation strings, not reliable lifecycle data. Please pass the actual status through the tool-end event and add regression coverage for these cases.



The cancellation path that returns after `tool_start` without a terminal `tool_end` predates this PR, including on main, so I am not asking you to own that behavior broadly. However, since this PR introduces the Chat lifecycle vocabulary, emitting `tool_end(status="cancelled")` for already-started tools would make that state actually reachable. 



Separately, `test_agent_tool_span_nests_under_agent_node_span` patches `_execute_node_inner` and calls the helper directly, so it does not exercise `LLMService.execute_with_tools` across the `asyncio.run` and `to_thread` boundaries. This is not blocking, but it would be good to close eventually.



Once the two items above are addressed, I think this will be ready to merge.

eryue0220:
> Thanks for the follow-up. The generated-file preservation, HITL reconciliation, fair trace allocation, metrics cleanup, Debug payload, and OTel wiring now look good.
> 
> I found two remaining issues before I can approve.
> 
> 1. Please restore cancellation checks after each completed tool.
> 
> `should_abort()` now runs only while a HITL pause is pending, so a cancellation that arrives after a normal tool completes is not observed until the next outer iteration. If the model returned several tool calls in one turn, later HTTP, MCP, or skill tools can still run after the user pressed Stop.
> 
> The fix should preserve the completed tool entry and its successful result, then stop the loop before starting the next tool. Please also make sure `_build_error_result` does not rewrite that completed entry as cancelled. `classify_tool_failure_status("Workflow execution cancelled", explicit_status="success")` still resolves to `cancelled`, so simply restoring the per-tool check would otherwise reintroduce the result-loss bug.
> 
> Please add a regression test that verifies both that the completed result remains successful and that a later tool in the same turn is not invoked.
> 
> 2. Please derive Chat lifecycle status from the structured tool result or an explicit status, rather than from the display summary.
> 
> At the current head, 20 production call sites invoke `_tool_end_yield`, and only two pass an explicit status. The summary heuristic is therefore the normal path, not an exception. For example:
> 
> ```
> 
> "Created task 'Fix request timeout' on board 'Infra'" -> timeout
> 
> "Card 'Cancelled deployment': active, 2 comment(s), 1 run(s)" -> cancelled
> 
> "Card 'Foo': error, 0 comment(s), 0 run(s)" -> success
> ```
> 
> These are presentation strings, not reliable lifecycle data. Please pass the actual status through the tool-end event and add regression coverage for these cases.
> 
> The cancellation path that returns after `tool_start` without a terminal `tool_end` predates this PR, including on main, so I am not asking you to own that behavior broadly. However, since this PR introduces the Chat lifecycle vocabulary, emitting `tool_end(status="cancelled")` for already-started tools would make that state actually reachable.
> 
> Separately, `test_agent_tool_span_nests_under_agent_node_span` patches `_execute_node_inner` and calls the helper directly, so it does not exercise `LLMService.execute_with_tools` across the `asyncio.run` and `to_thread` boundaries. This is not blocking, but it would be good to close eventually.
> 
> Once the two items above are addressed, I think this will be ready to merge.

Update for the remaining problem, thank you.

mbakgun:
Thanks @eryue0220  🙏   this round is much closer. The `run_async` OTel context fix was a good catch, and moving Chat away from summary-based status derivation is a clear improvement over main. 🚀 

One blocking issue remains: free-form error text can still override a structured tool status.

For example, this currently becomes `timeout`, even though the result explicitly says `error`:

```python
classify_tool_failure_status(
    "Variable 'request timeout' not found",
    explicit_status="error",
)
```

The same issue reaches `LLMService._tool_result_status`, so it can affect Chat, tool metrics, and tracing. In `ChatToolCall.vue`, an error uses destructive styling and a warning icon, while a cancelled call is muted with a Ban icon. A real failure can therefore look like a benign cancellation.

Please make a structured, recognized status win when it describes the same tool outcome. If a structured result only has an `error` string, default to `error` rather than inferring timeout or cancellation from user-controlled text. Please add regression coverage for the Chat helper and `_tool_result_status` using `request timeout` and `Cancelled deployment` in error text.

One important caveat: please do not make explicit status unconditional precedence inside `classify_tool_failure_status`. In `_run_one_tool`'s abort path, a cancelled `HumanReviewPause` can carry a derived `success` status even though the run was cancelled. That case must remain `cancelled`. Applying the precedence rule at the structured-result boundary, or excluding that abort call site, should avoid that regression.

Separately, and not blocking: there is still a narrow cancellation window between one tool's result event and the next sequential tool starting. This predates the PR, main has the same structure, so I am not asking you to own it here. A preflight `should_abort()` check at the start of `_run_one_tool` would close it cleanly later.

Once the status-classification fix is in, I am happy to approve. 🥇

eryue0220:
> Thanks @eryue0220 🙏 this round is much closer. The `run_async` OTel context fix was a good catch, and moving Chat away from summary-based status derivation is a clear improvement over main. 🚀
> 
> One blocking issue remains: free-form error text can still override a structured tool status.
> 
> For example, this currently becomes `timeout`, even though the result explicitly says `error`:
> 
> ```python
> classify_tool_failure_status(
>     "Variable 'request timeout' not found",
>     explicit_status="error",
> )
> ```
> 
> The same issue reaches `LLMService._tool_result_status`, so it can affect Chat, tool metrics, and tracing. In `ChatToolCall.vue`, an error uses destructive styling and a warning icon, while a cancelled call is muted with a Ban icon. A real failure can therefore look like a benign cancellation.
> 
> Please make a structured, recognized status win when it describes the same tool outcome. If a structured result only has an `error` string, default to `error` rather than inferring timeout or cancellation from user-controlled text. Please add regression coverage for the Chat helper and `_tool_result_status` using `request timeout` and `Cancelled deployment` in error text.
> 
> One important caveat: please do not make explicit status unconditional precedence inside `classify_tool_failure_status`. In `_run_one_tool`'s abort path, a cancelled `HumanReviewPause` can carry a derived `success` status even though the run was cancelled. That case must remain `cancelled`. Applying the precedence rule at the structured-result boundary, or excluding that abort call site, should avoid that regression.
> 
> Separately, and not blocking: there is still a narrow cancellation window between one tool's result event and the next sequential tool starting. This predates the PR, main has the same structure, so I am not asking you to own it here. A preflight `should_abort()` check at the start of `_run_one_tool` would close it cleanly later.
> 
> Once the status-classification fix is in, I am happy to approve. 🥇



> Separately, and not blocking: there is still a narrow cancellation window between one tool's result event and the next sequential tool starting. This predates the PR, main has the same structure, so I am not asking you to own it here. A preflight `should_abort()` check at the start of `_run_one_tool` would close it cleanly later.

Okay, update for previous. Thanks for your review.

mbakgun:
Thanks, `dc75370f` fixes the main status-classification issue well. Scoping the change to structured-result call sites was the right choice, and it keeps cancelled HITL pauses correct.

With Stop simulated from the first tool's `phase="result"` callback:

```text
main:    invoked = ['child']
PR head: invoked = ['child', 'later-child']
```

A user can now see the first result, press Stop, and still have the next sequential tool run. Please move the completed-tool abort check back after the terminal callbacks, and add a preflight check before the next tool emits `start` or reaches the executor.

The entry rewrite protection you added stays intact either way. `completed_tool_abort_reason` and the `_build_error_result` guard do not depend on where the check runs, so moving it back does not reintroduce result loss.

The existing test does not model this ordering. It flips the abort flag on a call counter, so the reason lands on the first tool's own check instead of after its result event, which is why CI stays green. Please test that the first entry remains `success` and the second tool never starts.

Two smaller status issues remain:

1. `{"error": {"message": "..."}}` is `error` in Chat but `success` in the Agent path because `_tool_result_error` only accepts strings. This appears latent rather than currently produced by a built-in tool, but identical input should not produce different lifecycle states. A non-null structured error should default to `error` when no recognized status exists.

2. `summarize_tool_calls()` still reclassifies statusless legacy entries from `result.error` text and ignores `result.status`. Please honor `result.status` first, then default to `error`.

For context, Ruff is clean, backend tests pass, frontend typecheck and lint are clean, and CI is green. None of that catches the ordering issue.

eryue0220:
> Thanks, `dc75370f` fixes the main status-classification issue well. Scoping the change to structured-result call sites was the right choice, and it keeps cancelled HITL pauses correct.
> 
> With Stop simulated from the first tool's `phase="result"` callback:
> 
> ```
> main:    invoked = ['child']
> PR head: invoked = ['child', 'later-child']
> ```
> 
> A user can now see the first result, press Stop, and still have the next sequential tool run. Please move the completed-tool abort check back after the terminal callbacks, and add a preflight check before the next tool emits `start` or reaches the executor.
> 
> The entry rewrite protection you added stays intact either way. `completed_tool_abort_reason` and the `_build_error_result` guard do not depend on where the check runs, so moving it back does not reintroduce result loss.
> 
> The existing test does not model this ordering. It flips the abort flag on a call counter, so the reason lands on the first tool's own check instead of after its result event, which is why CI stays green. Please test that the first entry remains `success` and the second tool never starts.
> 
> Two smaller status issues remain:
> 
> 1. `{"error": {"message": "..."}}` is `error` in Chat but `success` in the Agent path because `_tool_result_error` only accepts strings. This appears latent rather than currently produced by a built-in tool, but identical input should not produce different lifecycle states. A non-null structured error should default to `error` when no recognized status exists.
> 2. `summarize_tool_calls()` still reclassifies statusless legacy entries from `result.error` text and ignores `result.status`. Please honor `result.status` first, then default to `error`.
> 
> For context, Ruff is clean, backend tests pass, frontend typecheck and lint are clean, and CI is green. None of that catches the ordering issue.

Fix previous problems. Thanks you

mbakgun:
Thanks, `613dce76` resolves the three items from the previous round. I verified the Stop ordering after the first tool result, structured error handling on the Agent path, and `summarize_tool_calls` honoring `result.status`.

One blocking cancellation-boundary issue remains. Correcting earlier note: text inference is not safe on the exception path either, because generic exceptions can carry user-controlled text.

1. A returned tool error can cancel the whole Agent from free-form text alone.

Both `{"error": "Variable 'Workflow execution cancelled' not found"}` and `{"error": {"message": "Variable 'Workflow execution cancelled' not found"}}` produce a cancelled tool call and abort the loop. This is live rather than theoretical, since `_extract_tool_result` in `mcp_tool_executor.py:400` returns server-controlled error text directly in the tool result.

2. Thrown exceptions have the same problem. At `llm_service.py:1492`, `classify_tool_failure_status(exc)` turns a plain `ValueError` into a cancellation:

```text
ValueError("Variable 'Workflow execution cancelled' not found")
  -> tool_call.status = cancelled
  -> tool_metrics.cancelled = 1
  -> agent loop aborts
```

Any Python tool can raise that shape.

3. In the other direction, a real independently cancelled sub-workflow is swallowed. `_execute_sub_workflow_tool` catches `WorkflowCancelledError` in its generic handler at `workflow_executor.py:4082` and returns `{"status": "error", "error": "Workflow execution cancelled"}`. The new recognized non-cancel status rule then returns `None`:

```text
main:    error = "Workflow execution cancelled"   (agent stops)
PR head: error = None, text = "done anyway"       (agent continues)
```

This is reachable because sub-workflows get their own cancel event through `_register_sub_execution`. Parent cancellation bridges down, but not up, so `should_abort()` never fires when only the child is cancelled.

Please fix these as one boundary change:

- Returned tool payloads should abort only on explicit `status: "cancelled"`, never on `error` text.
- Preserve lifecycle status at the source. In `_execute_sub_workflow_tool`, catch `WorkflowTimeoutError` before `WorkflowCancelledError` and return explicit `timeout` or `cancelled` status.
- At the tool exception boundary, classify only trusted exception types: `WorkflowTimeoutError`, `WorkflowCancelledError`, and built-in `TimeoutError`. Generic exceptions should remain `error`.
- One tradeoff to decide deliberately: `codex_runner_service` and `opencode_runner_service` raise `ValueError("Command timed out after ...")`, which currently becomes `timeout`. Raising `TimeoutError` there would preserve that fidelity.
- Keep `should_abort()` as the dedicated trusted cancellation signal.

Please add regressions for an independently cancelled sub-workflow, explicit `status: "cancelled"`, returned string and structured errors containing the phrase, and a generic `ValueError` containing the phrase.

P2, fine as a follow-up:

- `_tool_result_status({"status": "", "error": "boom"})` returns success because an empty string normalizes to success before the error is considered.
- `summarize_tool_calls` reads `result.error` but not `result.outputs.error`, so `{"result": {"outputs": {"error": "child failed"}}}` counts as success while `_tool_result_status` says error.
- `normalize_tool_call_status("canceled")` returns unknown, although `text_indicates_cancellation` accepts both spellings.
- In the parallel sub-agent abort path, the first aborting entry is appended last while later siblings are committed first, so persisted `tool_calls` order does not match slot order.

The current tests cover a statusless structured cancellation message, but not the `status: "error"` payload produced by sub-workflow cancellation or the generic exception case.

Once this is covered, I am happy to approve.

eryue0220:
> Thanks, `613dce76` resolves the three items from the previous round. I verified the Stop ordering after the first tool result, structured error handling on the Agent path, and `summarize_tool_calls` honoring `result.status`.
> 
> One blocking cancellation-boundary issue remains. Correcting earlier note: text inference is not safe on the exception path either, because generic exceptions can carry user-controlled text.
> 
> 1. A returned tool error can cancel the whole Agent from free-form text alone.
> 
> Both `{"error": "Variable 'Workflow execution cancelled' not found"}` and `{"error": {"message": "Variable 'Workflow execution cancelled' not found"}}` produce a cancelled tool call and abort the loop. This is live rather than theoretical, since `_extract_tool_result` in `mcp_tool_executor.py:400` returns server-controlled error text directly in the tool result.
> 
> 2. Thrown exceptions have the same problem. At `llm_service.py:1492`, `classify_tool_failure_status(exc)` turns a plain `ValueError` into a cancellation:
> 
> ```
> ValueError("Variable 'Workflow execution cancelled' not found")
>   -> tool_call.status = cancelled
>   -> tool_metrics.cancelled = 1
>   -> agent loop aborts
> ```
> 
> Any Python tool can raise that shape.
> 
> 3. In the other direction, a real independently cancelled sub-workflow is swallowed. `_execute_sub_workflow_tool` catches `WorkflowCancelledError` in its generic handler at `workflow_executor.py:4082` and returns `{"status": "error", "error": "Workflow execution cancelled"}`. The new recognized non-cancel status rule then returns `None`:
> 
> ```
> main:    error = "Workflow execution cancelled"   (agent stops)
> PR head: error = None, text = "done anyway"       (agent continues)
> ```
> 
> This is reachable because sub-workflows get their own cancel event through `_register_sub_execution`. Parent cancellation bridges down, but not up, so `should_abort()` never fires when only the child is cancelled.
> 
> Please fix these as one boundary change:
> 
> * Returned tool payloads should abort only on explicit `status: "cancelled"`, never on `error` text.
> * Preserve lifecycle status at the source. In `_execute_sub_workflow_tool`, catch `WorkflowTimeoutError` before `WorkflowCancelledError` and return explicit `timeout` or `cancelled` status.
> * At the tool exception boundary, classify only trusted exception types: `WorkflowTimeoutError`, `WorkflowCancelledError`, and built-in `TimeoutError`. Generic exceptions should remain `error`.
> * One tradeoff to decide deliberately: `codex_runner_service` and `opencode_runner_service` raise `ValueError("Command timed out after ...")`, which currently becomes `timeout`. Raising `TimeoutError` there would preserve that fidelity.
> * Keep `should_abort()` as the dedicated trusted cancellation signal.
> 
> Please add regressions for an independently cancelled sub-workflow, explicit `status: "cancelled"`, returned string and structured errors containing the phrase, and a generic `ValueError` containing the phrase.
> 
> P2, fine as a follow-up:
> 
> * `_tool_result_status({"status": "", "error": "boom"})` returns success because an empty string normalizes to success before the error is considered.
> * `summarize_tool_calls` reads `result.error` but not `result.outputs.error`, so `{"result": {"outputs": {"error": "child failed"}}}` counts as success while `_tool_result_status` says error.
> * `normalize_tool_call_status("canceled")` returns unknown, although `text_indicates_cancellation` accepts both spellings.
> * In the parallel sub-agent abort path, the first aborting entry is appended last while later siblings are committed first, so persisted `tool_calls` order does not match slot order.
> 
> The current tests cover a statusless structured cancellation message, but not the `status: "error"` payload produced by sub-workflow cancellation or the generic exception case.
> 
> Once this is covered, I am happy to approve.

Thanks for the review. Addressed the remaining issues:

* Tool payloads abort only on explicit status: "cancelled", never from error text.
* Exception classification trusts only TimeoutError / WorkflowTimeoutError / WorkflowCancelledError; generic exceptions stay error.
* Sub-workflow/sub-agent now return explicit cancelled/timeout at the source; NodeResult stays success/error, lifecycle lives on the tool payload.
* Also fixed empty-status handling, outputs.error in metrics, canceled spelling, parallel abort ordering, and prior-timeout mislabeling.

mbakgun (approved):
Thanks for working through this so thoroughly @eryue0220  🙏  The cancellation boundary is now in the right place, the false-positive cases are covered, and real sub-workflow cancellation is preserved 💯

I also appreciate the focused regression coverage and the cleanup of the smaller status inconsistencies. I am approving and merging this. We can track the remaining timeout fidelity nuance and a few narrow edge cases as follow-ups, but they should not hold up this PR 🚀

PR #410, opened 2026-07-24
feat: add optional You.com search integration

mouse-value-add:
Adds a new You.com Search node to Heym workflows for web search capabilities. Supports keyless tier (100 searches/day) and API key for higher quotas. Complete docs and tests included.

### DIFF
--- a/backend/app/services/node_execution/nodes/youcom_search_node.py
+++ b/backend/app/services/node_execution/nodes/youcom_search_node.py
@@ -1,0 +1,168 @@
+"""You.com web search node for Heym workflows."""
+
+from __future__ import annotations
+
+import os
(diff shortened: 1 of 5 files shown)


mbakgun:
Hi thank you for your effort 🙏 This implementation feels incomplete. It should have included capabilities such as completing the credentials and updating the workflow DSL. Regarding Agents.md guidelines.

More importantly, this would be better suited as a template. You can submit it through the link below and share you.com's search integration with Heym users and people looking for templates.

Since it performs a single operation and depends on only one provider, it makes more sense as a template than as a standalone node. You can submit it here:

https://github.com/heymrun/heym#-share-a-template

PR #436, opened 2026-08-05
fix: explicitly release advisory lock before closing leader session

HassanRady:
_release_leader_session() previously only closed the SQLAlchemy session, returning the underlying connection to the pool with the pg_advisory_lock still held. This caused:

1. Self-lockout on error recovery: when the election loop hit an exception, the lock leaked onto the pooled connection and subsequent acquire attempts on a different pooled connection would fail permanently.

```Python
# Line 83-85
except Exception as e:
    logger.exception("Error in leader election loop: %s", e)
    await self._release_leader_session()  # ← lock leaks onto pooled connection
# Loop continues running, process stays alive
```
Here the process doesn't exit. The lock is stuck on a pooled connection, and when _try_acquire_leader_lock runs again 5 seconds later, it creates a new session which may get a different pooled connection. That new connection tries pg_try_advisory_lock and:

If it gets the same physical connection → succeeds (PG counts it as a re-acquire by the same session)
If it gets a different physical connection → fails, because the old connection in the pool still holds the lock

2. Delayed leader handoff on graceful shutdown: the lock persisted until process exit rather than releasing immediately on stop().

Now explicitly calls pg_advisory_unlock before session.close() so the lock is released deterministically.

### DIFF
--- a/backend/app/services/distributed_lock.py
+++ b/backend/app/services/distributed_lock.py
@@ -127,14 +127,7 @@ class DistributedLockService:
         if not self._leader_conn:
             return False
         try:
-            await self._leader_conn.execute(
-                text("SELECT pg_advisory_lock_shared(:lock_id)"),
-                {"lock_id": self._leader_lock_id},
-            )
-            await self._leader_conn.execute(
-                text("SELECT pg_advisory_unlock_shared(:lock_id)"),
-                {"lock_id": self._leader_lock_id},
-            )
+            await self._leader_conn.execute(text("SELECT 1"))
             return True
         except Exception:
             return False
--- a/backend/app/services/distributed_lock.py
+++ b/backend/app/services/distributed_lock.py
@@ -53,6 +53,13 @@ class DistributedLockService:
     async def _release_leader_session(self) -> None:
         if self._leader_session:
             try:
+                await self._leader_session.execute(
+                    text("SELECT pg_advisory_unlock(:lock_id)"),
+                    {"lock_id": self._leader_lock_id},
+                )
+            except Exception as e:
+                logger.warning("Failed to unlock advisory lock on release: %s", e)
+            try:
                 await self._leader_session.close()
             except Exception:
                 pass
--- a/backend/app/services/distributed_lock.py
+++ b/backend/app/services/distributed_lock.py
@@ -50,7 +50,7 @@ class DistributedLockService:
         await self._release_leader_session()
         logger.info("Distributed lock service stopped")
 
-    async def _release_leader_session(self) -> None:
+    async def _release_leader_session(self, force_invalidate: bool = False) -> None:
         if self._leader_conn:
             unlock_confirmed = False
             if self._is_leader:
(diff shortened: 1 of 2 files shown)


ckakgun:
Thanks for the contribution @HassanRady 🙏  The bug is real and the diagnosis is correct. I confirmed that the advisory lock survives `session.close()` and prevents other workers from becoming leader.

A few changes are needed before merging:

1. The aborted transaction path can still leak the lock. Calling `rollback()` first is not sufficient because `AsyncSession.rollback()` returns the connection to the pool, causing the unlock to run on a different PostgreSQL backend. Please keep the leader lock on a dedicated `AsyncConnection`, and invalidate the connection if the unlock cannot be confirmed.

2. Followers should not attempt to unlock. Right now every worker that fails to acquire runs an unlock it never owned, and Postgres logs `WARNING: you don't own a lock of type ExclusiveLock` for each one. Please track lock ownership explicitly, and avoid `pg_advisory_unlock_all()` since the pooled connection may hold unrelated advisory locks.

3. Please add focused tests for successful release, failed acquisition, and unlock failure.

Happy to merge once these are addressed 💯

HassanRady:
Thanks, @ckakgun, for noticing that. I have updated the PR.

ckakgun:
Thanks for the update. The main release path looks much better now, and the dedicated `AsyncConnection`, `AUTOCOMMIT`, and follower cleanup changes work as expected.

Two items still need attention before merging:

1. If the lock is acquired server-side but the client sees an error or cancellation, `_is_leader` remains `False` and the connection is returned to the pool with the lock still held. I reproduced this against a live Postgres. Whether the next worker can become leader depends on which pooled connection it gets, so the issue is easy to miss with an idle pool. Please invalidate the connection whenever the acquire result is uncertain.
2. Please add focused tests for successful release, follower cleanup, failed unlock, and uncertain acquire or cancellation.

It would also be good to clean up the connection if `execution_options()` fails after `engine.connect()` succeeds.

Thanks again for working through the feedback.

HassanRady:
Thanks, @ckakgun, for your feedback. I have updated the PR.

ckakgun:
Thanks, this looks good now. I checked the new commit against a live Postgres and the uncertain acquire no longer leaks the lock, the connection is invalidated as expected, `execution_options()` failures clean up the raw connection, and followers stay quiet. Re-raising `CancelledError` instead of swallowing it is a nice touch.

One thing left, and it is on the test file rather than the fix. The tests do not run in this repo, so all seven fail locally with "async def functions are not natively supported". `pytest-asyncio` is not one of our dependencies, so `@pytest.mark.asyncio` has no plugin behind it. This is not something CI approval will resolve, it will fail on the runner too.

Could you convert them to `unittest.IsolatedAsyncioTestCase`? That is what `AGENTS.md` asks for and it needs no new dependency. The scenarios you covered are the right ones, only the test/harness needs changing. Please also run `ruff format`, the file is missing a blank line before the last test. 

Once that is in I will approve and get this merged.

ckakgun:
I ran CI and expect the tests are failing.

HassanRady:
Thanks, ckakgun, for your feedback. I have updated the PR.

ckakgun (approved):
Thanks for the contribution and for working through the feedback. Everything looks good now. If you find Heym useful, we would really appreciate a star on the repository! ⭐

PR #439, opened 2026-08-05
feat: add cal-com trigger node

eryue0220:
This pr is to add a new node called `calCom Trigger` node, which is to support some of [cal.com](https://cal.com/docs/api-reference/v2/introduction) webhook API, and this pr is only focus on v2 API.


ckakgun:
Hi @eryue0220,

First, thank you. This is a serious piece of work: a backend service and node
handler, a credential flow with its own migration, docs, node registration,
Playwright coverage and real unit tests. You followed AGENTS.md more closely
than most PRs we receive. Nothing below is about the quality of the code.

We are closing this one, and I would rather give you the real reasoning than a
one line answer.

Cal.com maintains an official MCP server (https://cal.com/docs/mcp-server), and
Heym already has an MCP Call node plus MCP tools on the Agent node. Users can
book, reschedule and cancel today with their own credentials, and Cal.com keeps
that surface up to date instead of us. For a focused product like scheduling, we
would rather point people at the vendor's MCP server and ship a good template on
top of it than carry a vendor client in core. That is not a blanket rule against
vendor triggers, we ship several already, it is a call about this specific case.

If you are up for it, a template built on the Cal.com MCP server would be very
welcome at https://heym.run/submit. It reaches users in days instead of sitting
in review, and it carries your name.

Also, these files are unrelated to the Cal.com node:

- `backend/app/services/execution_recovery.py`
- `backend/app/services/execution_cancellation.py`
- `backend/app/services/ssrf_guard.py`

Last thing: we talk on Discord regularly. Please ping us before starting
something at this scale, since fifteen minutes of alignment would have saved you
a lot of hours. That is a gap in our process, not a mistake in your judgment.

Closing for now, not as a permanent no. Thanks again.

PR #476, opened 2026-08-20
Remove unused dependencies and dead helpers

uadhran:
Removes a few things that are not used anywhere.

- `class-variance-authority` — never imported
- `@types/marked` — `marked` already ships its own types
- `mako` as a direct backend dependency — Alembic still pulls it in
- `hasNestedDollarInParentheses` and `hasMultipleDollarExpressions` in `frontend/src/lib/utils.ts` — nothing calls them

Lint, typecheck, and the backend test suite passed locally.


ckakgun:
Thanks for this @uadhran , the cleanup is welcome 🚀 

Three of the four are good and I'll take them. @types/marked is a no-op stub (marked@18 ships its own types, and dropping it also removes a duplicate marked@17). class-variance-authority isn't imported anywhere, though heads up that shadcn-vue generated components use it, so it may come back later. The two $ helpers have never had a call site in any revision, and the real logic lives in the backend executor, so those are correct removals.

One change please: keep mako. It isn't a leftover import, it's a version floor. It was added deliberately in 5f005b06 to move us to 1.3.12, and Alembic's metadata is just Requires-Dist: Mako with no lower bound, so this line is the only thing holding it.

CVE-2026-41205, affects <= 1.3.10, fixed in 1.3.11
CVE-2026-44307, affects <= 1.3.11, fixed in 1.3.12
Nothing breaks today, the lock still resolves 1.3.12. But you did find a stale pin, so the fix is to raise it rather than drop it:


"mako>=1.3.12",
Also worth knowing for future sweeps: skills run on the backend venv and can't install their own packages, so pyproject.toml is their whole import surface. reportlab and python-docx are never imported under app/ but are advertised to users in skill_builder.py. "Not imported" alone isn't a safe removal test on the backend. mako is fine there since Alembic keeps it installed.

Happy to merge once that line is back at >=1.3.12. Thanks again. 🙏

uadhran:
Kept mako and raised the floor to `>=1.3.12`. Alembic still does not pin a minimum, so that line is what stops an older Mako from coming in.

The other three removals are unchanged.

ckakgun (approved):
Thanks @uadhran  🌸

PR #477, opened 2026-08-20
Add ENCRYPTION_KEY to the README env table

uadhran:
The short environment table lists `SECRET_KEY` but not `ENCRYPTION_KEY`. The backend will not start without both.

Adds the missing row, with the same generator already documented in ENVIRONMENT-VARIABLES.md.


ckakgun (approved):
Thank you 💯

PR #478, opened 2026-08-20
refactor: use shared node icons on the canvas and palette

uadhran:
The node palette already used `lib/nodeIcons.ts` for template rows, then kept a second Lucide map for builtin cards. A new node type could get the right icon in one place and the wrong one in the other.

Builtin cards now use the shared map. The duplicate local map is gone. A small test checks every node definition has an icon and a color class.

### DIFF
--- a/frontend/src/components/Nodes/BaseNode.vue
+++ b/frontend/src/components/Nodes/BaseNode.vue
@@ -1,13 +1,12 @@
 <script setup lang="ts">
 import { computed } from "vue";
 import { Handle, Position, useVueFlow } from "@vue-flow/core";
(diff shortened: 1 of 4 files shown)


ckakgun:
Thanks again 🌸  and for catching the duplication.

I compared the maps: all 60 node types currently resolve to the same icons, so this is a behavior-preserving change. I also ran the branch locally: typecheck, lint, the file-line-limit check, and the new test all pass. `NodePanel` no longer retains its local map, and the import cleanup is correct.

One request 🙏  before merge: `NodePanel` is not the only remaining icon map.

- `frontend/src/components/Nodes/BaseNode.vue:26` still owns a local `icons` map used by the primary canvas surface. It already consumes `nodeIconColorClass` from `lib/nodeIcons`, so it is only partially migrated.
- `frontend/src/components/Panels/propertiesPanel/usePropertiesPanelController.ts:96` redeclares the same 60-entry `nodeIcons` map inside the composable.

Both maps match `lib/nodeIcons` today, so nothing is currently broken. However, the underlying drift risk described in this PR remains, especially on the canvas, which is a more visible surface than the palette. Could you import and use the shared `nodeIcons` map in both places as part of this cleanup?

Non-blocking: the added test is mostly redundant with the `Record<NodeType, ...>` compile-time guarantee, and #479 addresses getting Vitest tests into the standard check/CI path. I would also consider `refactor` a slightly more precise title than `fix`, since this prevents a future visual regression rather than correcting a currently visible mismatch. 🚀

uadhran:
BaseNode and the properties controller now import `nodeIcons` from `lib/nodeIcons.ts`. The local 60-entry maps are gone. `nodeColorMap` is unchanged — that is CSS token names, not the Lucide map.

Title set to refactor. Lint, typecheck, and 116 Vitest tests passed locally.

ckakgun:
Thank you @uadhran 💯

PR #479, opened 2026-08-20
Run frontend unit tests in check.sh and CI

uadhran:
There are already 16 Vitest files under frontend/src. They never ran: no `test` script, and neither check.sh nor PR checks invoked them.

Adds `bun run test` (`vitest run`) and calls it from check.sh and the PR check job. Locally: 16 files, 115 tests passed.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -39,6 +39,10 @@ jobs:
         working-directory: frontend
         run: bun run typecheck
 
+      - name: Frontend unit tests
+        working-directory: frontend
+        run: bun run test
+
       - name: Frontend build
         working-directory: frontend
         run: bun run build
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,7 +9,7 @@ Read and follow this `AGENTS.md` at the start of every session. Repository conve
 ```bash
 ./run.sh                    # Start all services (postgres, backend, frontend)
 ./run.sh --no-debug         # Start with INFO logging instead of DEBUG
-./check.sh                  # Run frontend lint/typecheck, backend Ruff checks, and backend tests
+./check.sh                  # Run frontend lint/typecheck/unit tests, backend Ruff checks, and backend tests
 ./run_e2e.sh                # Run frontend Playwright E2E tests separately
 SECRET_KEY=test-secret-key-for-tests-only-32-bytes ./check.sh  # Use when no SECRET_KEY is exported locally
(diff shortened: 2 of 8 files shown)


ckakgun (changes requested):
Thanks for this @uadhran . The gap is real and the implementation is clean. Both CI jobs are green, and I verified the branch locally as well: lint and typecheck pass, and the unit suite completes 16 files and 115 tests in under a second 💯 

I also checked why `vitest.config.ts` is necessary. A bare `vitest run` collects the 32 Playwright specs under `frontend/e2e/`, which are not valid Vitest suites and fail during collection. The `include` scope is therefore essential.

Please keep `bun run test` out of `check.sh` and run it only in the CI `check` job. We want to preserve the existing scope of the fast local command: frontend lint and typecheck, plus backend checks and tests. CI should enforce this new frontend unit-test gate.

With that removed, revert the `./check.sh` Quick Start description in `AGENTS.md` to its original wording: `Run frontend lint/typecheck, backend Ruff checks, and backend tests.` Keep the new standalone `bun run test` command under the Frontend section, but describe it as CI-enforced rather than as a required local pre-commit command.

Please also update the documentation:

1. `README.md:381`: add the frontend Vitest unit suite to the PR-check list.
2. `README.md:718`: add `bun run test` to the individual frontend validation command.
3. `AGENTS.md`: the Testing section and the `**Testing:**` guideline still describe frontend testing as Playwright-only. Document that Vitest specs live at `frontend/src/**/*.test.ts`, run with `bun run test`, and are enforced in CI.

Non-blocking: the current include pattern intentionally covers today's 16 `*.test.ts` files, but future `.spec.ts` or `.test.tsx` files would not be discovered. Also, `vitest.config.ts` is outside both TypeScript project includes, so it is not typechecked. Please consider adding it to `tsconfig.node.json`.

uadhran:
Took `bun run test` out of `check.sh`. Local check is lint, typecheck, and backend again. CI still runs the Vitest suite.

Quick Start wording for `./check.sh` is restored. `bun run test` is documented as CI-enforced. README PR-check list and the individual frontend command now include it. Testing notes in AGENTS.md cover `frontend/src/**/*.test.ts`.

`vitest.config.ts` is in `tsconfig.node.json`. Include is `src/**/*.{test,spec}.{ts,tsx}` so Playwright files under `e2e/` are still not collected.

ckakgun (approved):
Thank you so much !

PR #494, opened 2026-08-25
feat(frontend): summarize execution timeline health

gegemeimingzi:
## What changed
- Add a compact execution summary above the timeline with duration, span count, failures, retries, and a healthy state.
- Keep the summary derived from the complete timeline model so hiding rows does not change run health metrics.
- Add unit coverage for healthy/empty and failed/retried executions.

## Validation
- 
px vitest run src/components/Panels/executionTimeline.test.ts (5 passed)
- 
px vue-tsc --noEmit
- 
px eslint src/components/Panels/ExecutionTimeline.vue src/components/Panels/executionTimeline.ts src/components/Panels/executionTimeline.test.ts
- 
px vite build
- git diff --check

This is intentionally scoped to the execution timeline UX and does not change backend behavior or execution semantics.

### DIFF
--- a/frontend/src/components/Panels/ExecutionTimeline.vue
+++ b/frontend/src/components/Panels/ExecutionTimeline.vue
@@ -13,6 +13,7 @@ import {
   buildTimelineModel,
   formatTimelineMs,
   getTimelineRowKey,
+  summarizeTimelineModel,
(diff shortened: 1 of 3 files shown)


ckakgun (approved):
Thanks 🌸

PR #496, opened 2026-08-25
feat(frontend): add execution span details inspector

gegemeimingzi:
## Summary
- Add a collapsible, scrollable details section below the execution timeline when a span is selected.
- Show status, duration, attempts/retries, HITL wait state, last error, trace ID, trace navigation, and node output.
- Reuse the existing JsonTree component for object/array output previews and keep a text fallback for scalar output.
- Add a release-tour entry (disabled until release) and a Playwright interaction assertion.

## Validation
- Vitest: 26/26 passing (timeline + release-tour mapper)
- vue-tsc --noEmit: passing
- ESLint: passing
- Vite production build: passing
- git diff --check: passing
- Playwright scenario added; full E2E requires the repository PostgreSQL/UV service stack and will run in CI.

Implementation follows the direction agreed in Discussion #495: details below the timeline, JsonTree reuse, descriptive retry/HITL fields, and separate release-tour coverage.

Related: #494, Discussion #495


ckakgun:
Screenshot 2026-08-25 at 2 03 25 PM


Thanks for this, the direction is right and the panel is genuinely useful. A few things
before it can go in.

1. Output renders twice. The v-else attaches to v-if="selectedSpan.isHitlWait" instead of
   the object check above it, so for object outputs you get both the JsonTree and the raw
   JSON.stringify <pre>. Null outputs print a literal "null" too.

2. The E2E failure is deterministic, not flaky. The "-wait" suffix only appears when
   isHitlWait is true, which comes from metadata.hitl_wait on a HITL review resume or from
   agent/codex nodes. wait_live_two is a plain wait node, so the locator never resolves and
   the click hangs until the timeout. Assert against a normal span instead, and please drop
   the state from the test id.

3. The release tour entry never reaches the UI. With tourEnabled: false the whole release
   is dropped by buildReleaseTours, and that is the only consumer. Turning the flag on is
   enough, support.ts derives the seeded id from the registry. 

4. Could the details fill the panel instead of stacking under it? The scrollable rows list
   staying above is not needed, the X already closes it. Also please move the block after
   the rows in the template rather than pushing it down with order-10, that split breaks
   tab and screen reader order.

Smaller ones: formatOutput() runs from the template so the full output is re-stringified
every render during live runs, a computed would do. The span is still a div with no role
or keyboard path, and there is no highlight for the selected span. "Last error" shows on
spans that eventually succeeded with no hint it came from an earlier attempt.
retryMaxAttempts is plumbed through but never shown. error ?? retryLastError should be ||.
Docs are untouched and docTarget points at workflows-tab, which never mentions the
timeline. ExecutionTimeline.vue is at 679 lines now, so the details probably deserve their
own component.

Nice touch on the watch(rows) that re-resolves the selected span as the model refreshes.
Happy to take another look 🙏

ckakgun:
Thanks @gegemeimingzi  💯  most of it looks good now. A few changes left 🙏 

- Move ExecutionSpanDetails out of the 220px overflow-y-auto rows container, as a sibling
  of it. Inside it the panel opens below the fold on any run with ~9 or more rows, and it
  creates a nested scroll. It should take over the rows area when open.

- Drop class="order-10" from the component root. The parent is not flex anymore, so it
  does nothing.

- Move the new docs section above ## Related. Right now the related links list sits under
  the new heading.

- Import formatTimelineMs from executionTimeline.ts instead of the inline copy in
  ExecutionSpanDetails, so the two cannot drift.

- Remove the Wait field, or replace it with something informative like the wait duration.
  isHitlWait is only true for HITL review spans, so it shows "None" on virtually every
  span a user clicks. The E2E assertion on it should move to the node label (waitLiveTwo).

- Fix import order in the new component: internal types before internal code.

Everything else from the last round is resolved 👍

gegemeimingzi:
@ckakgun — all six items from your last round are addressed:

1. `ExecutionSpanDetails` moved out of the 220px rows container — it's now a sibling (`<template v-if>` + `<div v-else>` on the rows container), so it replaces the rows area when open; no nested scroll and no below-fold on long runs.
2. `order-10` dropped from the component root.
3. The docs section now sits above `## Related`.
4. `formatTimelineMs` is imported from `executionTimeline.ts`; the inline copy in `ExecutionSpanDetails` is gone.
5. The Wait field is removed, and the E2E assertion moved to the node label (`waitLiveTwo`) on the details panel.
6. Import order in the new component follows the repo rule (internal types before internal code).

I also changed how the timeline sits in the Execution Log: opening it now gives the timeline the full panel height (the node-results list is hidden while the timeline is open, and rows/details fill the space), and clicking the timer from a collapsed log expands the panel first. Previously the timeline was squeezed to roughly half the log. Let me know if you'd rather keep the log visible alongside the timeline.

ckakgun:
Thanks, the panel layout is right now. Two things left from testing it.

- The hover tooltip stays stuck on screen after clicking a span. Clicking swaps the rows
  out for the details panel, so the span unmounts and mouseleave never fires, leaving
  hoveredSpan set. Clearing it in onSpanClick should do it.

- There is a large empty area under the details. The Output block is still capped at
  max-h-44, so it stops at 176px while the panel now fills the full height. Letting it
  grow to fill the remaining space would use that room.



Screenshot 2026-08-25 at 8 04 20 PM

Screenshot 2026-08-25 at 8 04 16 PM




Taking over the rows on click and going to max height are both fine as they are.

3ni8ma:
@gegemeimingzi Nice PR — well-structured and the timeline → details flow makes sense. A few observations:

**Accessibility (good)**
- Keyboard support on spans (`onSpanKeydown` with Enter/Space) is solid
- `role="button"`, `tabindex="0"`, `aria-label`, and `aria-pressed` on span elements are well done
- Close button has both `title` and `aria-label`

**Potential issues**

1. **ExecutionSpanDetails.vue — `outputText` computed is unused in template logic**: The computed `outputText` is defined but only used in the `<pre>` fallback. Since `JsonTree` handles objects/arrays and `outputText` handles scalars + JSON-stringified objects, the `JSON.stringify` path for non-string objects is dead code — objects go through `JsonTree` via the `v-else-if` branch. Consider simplifying to just `String(span.output)` for the scalar fallback.

2. **`traceIdCopied` uses `window.setTimeout`**: This won't clean up if the component unmounts before the 1.5s timeout. Minor, but consider using `onUnmounted` or the Vue `onBeforeUnmount` lifecycle hook to clear it, or use a pattern like `useTimeoutFn` from VueUse.

3. **DebugPanel.vue — `toggleTimeline` auto-expands to `maxHeight`**: When the panel is collapsed and you click the timer, it opens to `maxHeight`. This might feel aggressive if the user has a small monitor. Consider using `Math.min(maxHeight, 500)` or similar to cap the initial expansion.

4. **`v-if` vs `v-show` for the debug panel content area**: The change from `v-if="!isCollapsed"` to `v-if="!isCollapsed && !showTimeline"` hides the content area entirely when the timeline is shown. This means switching from timeline back to content will remount the scroll container. If users frequently toggle between them, this could cause scroll position loss. `v-show` might be more appropriate here if that matters.

5. **`ExecutionSpanDetails` reactivity in `ExecutionTimeline.vue`**: The `watch` on `rows` that replaces `selectedSpan` is correct but could benefit from a brief comment explaining *why* — the selected span object comes from the previous `rows` computation, and when `rows` recomputes (e.g., new results arrive), the old span object would be stale.

**Nits**
- The `SpanDetailsInspectorTourVisual.vue` is nicely self-contained — good pattern for release tour visuals.
- The `formatTimelineMs` reuse is clean.
- E2E assertion is minimal but covers the happy path well.

**Overall**: This is a solid, well-tested feature. The items above are suggestions, not blockers. Happy to approve once the `v-if`/`v-show` decision for the content area is intentional.

gegemeimingzi:
Thanks @ckakgun and @3ni8ma for the detailed review — all feedback is addressed in `c0af542`:

**ckakgun's two items:**

1. **Stuck hover tooltip** — `onSpanClick` now clears `hoveredSpan` before opening the details panel. Since clicking swaps the rows out, the span unmounts and `mouseleave` never fires; clearing it in the click handler fixes the stuck tooltip.
2. **Empty area under the details** — the Output block dropped `max-h-44` and now uses `flex-1 min-h-0`, so it grows to fill the remaining panel height (still scrollable for long outputs).

**3ni8ma's suggestions:**

1. **`outputText` dead code** — simplified to `String(span.output)`; objects/arrays render through `JsonTree` via the `v-else-if` branch, so the `JSON.stringify` path was indeed unreachable for non-string values.
2. **`traceIdCopied` timeout cleanup** — the timer is now stored and cleared in `onBeforeUnmount` (and re-armed on repeat clicks rather than stacking).
3. **`toggleTimeline` initial expansion** — the collapsed-panel path now caps at `Math.min(maxHeight, 400)` instead of jumping straight to `maxHeight`; 400 matches the existing threshold in the non-collapsed path.
4. **`v-if` → `v-show` for the content area** — switched to `v-show` so toggling the timeline doesn't remount the scroll container and lose scroll position. This is the intended behavior: the scroll container persists, `scrollToBottom`/selection watchers are unaffected, and the ref stays stable.
5. **`watch(rows)` rationale** — added a comment explaining that `selectedSpan` is a snapshot from the previous `rows` computation and is re-resolved by key against the fresh rows (or closes if the span no longer exists).

All three files — `DebugPanel.vue`, `ExecutionSpanDetails.vue`, `ExecutionTimeline.vue` — pass lint, typecheck, and the 122 Vitest tests. Happy to adjust anything else 🙏

mbakgun (commented):
Thank you so much 😊

PR #500, opened 2026-08-26
refactor: move clickhouse node execution into its own module

fatihkaratash:
ClickHouse was the last node still executing inside `workflow_executor.py`. The other 61 already live in `node_execution/nodes/`, and `clickhouse_node.py` was just a wrapper calling back into the executor.

Moved the body of `_run_clickhouse_node` into `clickhouse_node.execute` and removed the method. Three tests in `test_clickhouse_service.py` called it directly, so they now build a `NodeExecutionContext` the same way `test_jira_node.py` does.

No behavior change. `workflow_executor.py` drops from 9604 to 9499 lines, which also leaves a bit more room under the 10000 limit in `check-file-line-limits.sh`.

Checked: `ruff check` and `ruff format --check` are clean, and `pytest tests/test_clickhouse_service.py tests/test_credential_clickhouse.py` gives 30 passed. My full-suite run has 21 failures, but I get the same 21 on a clean `main` — they're Docker and symlink tests failing on Windows, unrelated to this.

The README asks for a Discussion first. This one looked mechanical enough to just show the diff, but say the word and I'll move it.


ckakgun (approved):

ckakgun:
Thanks @fatihkaratash 🌸 !

PR #507, opened 2026-08-29
Bump qdrant-client to 1.19 and skip server compatibility check

heym-coder:
## Change Summary

Upgraded `qdrant-client` from `>=1.17.1` to `>=1.19.0` and disabled the client's server-version compatibility check on every `QdrantClient` construction site, so Heym's vector store connects reliably to Qdrant server versions the client would otherwise refuse (which previously surfaced as startup/connection failures against compatible servers).

Changes:
- `backend/pyproject.toml` / `backend/uv.lock` — bump `qdrant-client` to `>=1.19.0` (resolved to 1.19.0).
- `backend/app/services/qdrant_pool.py` — pass `check_compatibility=False` in both the API-key and no-API-key pool construction branches.
- `backend/app/services/vector_store.py` — pass `check_compatibility=False` in both `QdrantVectorStoreService` construction branches.

Verification: `uv lock --check` resolves cleanly, `ruff check` and `ruff format --check` pass on the touched modules, and `tests/test_rag_credential.py` (the suite covering `QdrantClient` construction) passes 41/41. Backend-only change, so no UI screenshot is required.


mbakgun:
duplicated

PR #525, opened 2026-09-03
Fix execution history deletion authorization

roonakyadav:
## Problem

Workflow collaborators can currently delete execution history belonging
to the workflow owner.

## Fix

- Restrict bulk execution-history deletion to workflows owned by the current user.
- Return `403` when a collaborator attempts per-workflow history deletion.
- Add regression tests for collaborator and owner behavior.

## Validation

- 31 related backend tests passed
- Ruff passed
- git diff --check passed

This implements the behavior requested by the maintainers following
verification of the reported issue.

### DIFF
--- a/backend/app/api/workflows.py
+++ b/backend/app/api/workflows.py
@@ -1135,17 +1135,10 @@ async def clear_all_execution_history(
     current_user: User = Depends(get_current_user),
     db: AsyncSession = Depends(get_db),
 ) -> None:
-    accessible_workflows = select(Workflow.id).where(
-        or_(
-            Workflow.owner_id == current_user.id,
-            Workflow.id.in_(
-                select(WorkflowShare.workflow_id).where(WorkflowShare.user_id == current_user.id)
-            ),
-        )
-    )
+    owned_workflows = select(Workflow.id).where(Workflow.owner_id == current_user.id)
     await db.execute(
         ExecutionHistory.__table__.delete().where(
-            ExecutionHistory.workflow_id.in_(accessible_workflows)
+            ExecutionHistory.workflow_id.in_(owned_workflows)
         )
     )
     await db.execute(RunHistory.__table__.delete().where(RunHistory.user_id == current_user.id))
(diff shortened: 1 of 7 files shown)


mbakgun (changes requested):
Hi @roonakyadav, thank you for the PR and the backend tests. The authorization changes look correct.

Before merging, please address the remaining UI issues:

- Hide **Clear history** for collaborators.
- Preserve history when the API returns `403`.
- Reload the global history list after bulk deletion because shared workflow history remains.
- Run `ruff format` and add frontend coverage for these behaviors.

The backend test suite passed locally. Once these points are addressed, the PR should be ready for another review. 👍

roonakyadav:
@mbakgun Thanks for the review! I've addressed all the remaining UI issues:

Hid the Clear history button for collaborators in the per-workflow dialog.

Preserved the local history cache when the API returns a 403 on a forbidden clear.

Updated the global all-history dialog to reload the list after a bulk deletion, ensuring shared workflow history correctly remains visible.

Formatted the backend changes with ruff format and added frontend unit tests covering these store behaviors.

Let me know if you need anything else or if this is good to go! 👍

mbakgun:
> @mbakgun Thanks for the review! I've addressed all the remaining UI issues:
> 
> Hid the Clear history button for collaborators in the per-workflow dialog.
> 
> Preserved the local history cache when the API returns a 403 on a forbidden clear.
> 
> Updated the global all-history dialog to reload the list after a bulk deletion, ensuring shared workflow history correctly remains visible.
> 
> Formatted the backend changes with ruff format and added frontend unit tests covering these store behaviors.
> 
> Let me know if you need anything else or if this is good to go! 👍

Thanks for the follow-up. The authorization and stale UI issues are fixed now, and the relevant checks pass locally. The only remaining point is the “Clear All” copy:

- `ExecutionHistoryAllDialog.vue`, `clearAllHistory()`: clarify that shared workflow history is retained.
- `docs/content/reference/execution-history.md`: update the “All History” and “Clear History” descriptions to match this behavior. 🙂

roonakyadav:
Just pushed the copy updates for both the confirmation dialog and the docs! The UI prompt and the execution-history.md now clearly explain that history for shared workflows is retained. Thanks again for the review! Appreciate your patience though

mbakgun (approved):

mbakgun:
Thank you so much 😊

PR #526, opened 2026-09-03
Audit history deletion and align All History with the access model

mbakgun:
Follow-up to #525. That PR closed the authorization hole; these are the three gaps left around it, plus the advisory acknowledgment.

## 1. The bulk clear left no audit trail

`DELETE /workflows/history/all` deletes every run of every workflow the caller owns. It emitted nothing, while the narrower per-workflow clear was audited. It now emits `workflow.history_clear_all` with the number of rows actually removed:

```
action=workflow.history_clear_all outcome=success actor_id=… actor_email=…
target=user:… workflow_runs_deleted=7 chat_runs_deleted=3
```

Recording the counts means the trail says how much was destroyed, not just that something was. The call sits after the deletes for that reason, matching `workflow.history_view`, which also audits after it knows its counts.

## 2. The new 403 left no trail either

A collaborator reaching for someone else's history got a 403 and vanished from the record. `delete_workflow` already audits that refusal with `outcome=denied` / `reason=not_owner`; the history clear now does the same, before raising:

```
action=workflow.history_clear outcome=denied reason=not_owner actor_id=… target=workflow:…
```

Both calls stay in the router, so the trail stays on the main instance.

## 3. All History disagreed with the access model

`list_all_execution_history` and `get_execution_history_entry` each carried their own copy of an owner-or-direct-share clause. `get_workflow_for_user` uses `workflow_access_clause`, which also covers team shares — so a workflow shared with your team showed its runs in the editor's history dialog but not in All History, and the two aggregated endpoints could only ever agree with each other by accident.

Both now call `workflow_access_clause`. This exposes nothing new: every row it adds belongs to a workflow the caller can already open, and whose per-workflow history endpoint already serves it. Two hand-rolled copies of the predicate are gone.

Deliberately **not** changed: `get_recent_history_for_user`, which feeds run outputs into dashboard chat context. Widening what an LLM prompt sees is a separate decision, so it keeps the narrower clause.

## 4. Test stubs were faking the audit path

The actor stubs were `SimpleNamespace(id=...)` with no `email`. `audit()` reads `actor.email`, throws `AttributeError`, and swallows it into a `logger.warning`. So the audit line silently failed to emit on every one of those tests while they stayed green.

The stubs now carry an email, and the new assertions read the emitted log line rather than trusting that the call happened. A broken audit call site now fails the suite instead of logging a warning nobody reads.

## 5. SECURITY.md

Acknowledgment for @fatihkaratash — GHSA-fmpw-hj3m-xvj6, the Redis node reading an inaccessible or missing credential as an empty configuration, so an authorization result fell through to the `localhost:6379` connection defaults instead of failing closed. Reported and fixed by the same person; the fix merged in 6a27cf1c.

## Tests

Five new cases in `test_execution_history_api.py`. All five were confirmed to fail against the pre-change code:

- `test_owner_clear_emits_a_success_audit_line`
- `test_denied_clear_is_audited_rather_than_silently_rejected`
- `test_bulk_clear_is_audited_with_the_number_of_rows_removed`
- `test_all_history_list_reaches_team_shared_workflows`
- `test_all_history_entry_reaches_team_shared_workflows`

The three audit tests also fail if the actor stub loses its `email`, which is what makes gap 4 stay closed.

## Validation

Run from a clean checkout with no local `.env`:

- `./run_tests.sh` — 3886 / 3886 passed
- `ruff format --check .` — 699 files unchanged
- `ruff check .` — passed
- `bun run lint:check`, `bun run typecheck`, `bun run test` (134), `bun run build` — passed
- `scripts/check-file-line-limits.sh` — passed

## Notes

- No release tour entry. The only user-visible effect is that team-shared runs now appear in All History, which is a correction to an existing inconsistency rather than a feature. Happy to add one if you'd rather announce it.
- `clear_workflow_versions` has the same owner check with no denied audit. Left alone to keep this change scoped; worth a follow-up if you want the two to match.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -98,3 +98,8 @@ vulnerabilities in Heym:
   egress guard, and contributing that fix (GHSA-6rph-qqcv-jqh4), and a cluster
   of capability secrets stored or returned in plaintext across webhook auth,
   MCP keys, portal sessions, and Discord interactions (GHSA-6x65-w7q7-wg93).
+- [@fatihkaratash](https://github.com/fatihkaratash) for the Redis node reading
+  an inaccessible or missing credential as an empty configuration, so an
+  authorization result fell through to the `localhost:6379` connection defaults
+  instead of failing closed (GHSA-fmpw-hj3m-xvj6), and for contributing the
+  remediation.
--- a/backend/app/api/workflows.py
+++ b/backend/app/api/workflows.py
@@ -945,16 +945,7 @@ async def get_execution_history_entry(
         select(ExecutionHistory, Workflow)
         .join(Workflow, ExecutionHistory.workflow_id == Workflow.id)
         .where(ExecutionHistory.id == entry_id)
-        .where(
-            or_(
-                Workflow.owner_id == current_user.id,
-                Workflow.id.in_(
-                    select(WorkflowShare.workflow_id).where(
-                        WorkflowShare.user_id == current_user.id
-                    )
-                ),
-            )
-        )
+        .where(workflow_access_clause(current_user.id))
     )
     row = exec_result.first()
     if row:
@@ -1034,16 +1025,7 @@ async def list_all_execution_history(
             ExecutionHistory.executed_by_instance_name,
         )
         .join(Workflow, ExecutionHistory.workflow_id == Workflow.id)
-        .where(
-            or_(
-                Workflow.owner_id == current_user.id,
-                Workflow.id.in_(
-                    select(WorkflowShare.workflow_id).where(
-                        WorkflowShare.user_id == current_user.id
-                    )
-                ),
-            )
-        )
+        .where(workflow_access_clause(current_user.id))
     )
     if workflow_id:
         exec_subq = exec_subq.where(ExecutionHistory.workflow_id == workflow_id)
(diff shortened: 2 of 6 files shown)


mbakgun:
Pushed `fe34d46c` for the red `check` job.

The two failures were `test_a_loop_keeps_dispatching_while_earlier_dispatches_run` and `test_emits_hidden_heartbeat_while_waiting_for_model_stream`. Neither is touched by this PR, and neither was touched by the Redis merge in `6a27cf1c` — both files last changed in `d4991a98` and `b5eab9c0`. They pass locally when the machine is idle and fail under the parallel load `run_tests.sh` creates, which is what makes them look like someone else's breakage.

Each asserted something the measurement cannot support:

**The loop test** gave the whole run an absolute budget of two 700ms waits. CI measured 1.52s against the 1.4s ceiling — a healthy run, just a loaded runner. Worse, the boundary it picked sits inside the noise: a real one-batch run and a hypothetical two-batch run are both ~1.4-1.5s, so the ceiling could not separate them even in principle. It now compares the work the run contains against the run's own wall clock. Ten overlapping dispatches hold ~7s of waiting inside a run that costs about one wait; serialized, the two numbers converge on each other. Serialization fails the new assertion arithmetically rather than by margin, and the deadlock case stays covered outright by the sibling test's 30s completion check.

**The heartbeat test** required the stream to end on the pull immediately after the final chunk. The payload and the terminating sentinel reach the consumer through two separate `call_soon_threadsafe` hops, so with the interval patched to 1ms a keepalive can land between them — which is what keepalives exist to do. It now drains trailing heartbeats and asserts what `_stream_sse_with_heartbeat` actually guarantees: nothing but heartbeats after the last chunk, and the stream ends. A stream that never ends still fails.

Verification: the two suites ran 5/5 green in isolation, and the full parallel suite ran 4 times consecutively at 3886/3886. Ruff format and check pass.

Happy to split these two into their own PR if you would rather keep this one on the audit change.

PR #535, opened 2026-09-04
fix: handle unknown workflow node types

roonakyadav:
## Summary

Fixes a mobile workflow editor crash when a persisted workflow contains an unknown, stale, or unsupported node type.

## What changed

- Added a shared safe node display resolver for mobile workflow surfaces.
- Added safe fallbacks for labels, summaries, icons, and icon colors.
- Preserved unknown nodes instead of dropping them.
- Guarded workflow edge handling when an unknown source node has no explicit source handle.
- Added focused tests for the resolver, mobile surfaces, and workflow edge behavior.

## Validation

- `npm test -- --run`
- `vue-tsc --noEmit`
- ESLint
- `git diff --check`
- `./check.sh`

The backend portion of `./check.sh` reaches 358 failures caused by missing `SECRET_KEY` / `ENCRYPTION_KEY` configuration; the same failures reproduce on the pre-change baseline.
  
Related: #534

### DIFF
--- a/frontend/src/components/Canvas/MobileWorkflowTree.vue
+++ b/frontend/src/components/Canvas/MobileWorkflowTree.vue
@@ -11,7 +11,7 @@ import MobileWorkflowTreeHeader from "@/components/Canvas/MobileWorkflowTreeHead
 import MobileWorkflowTreeMore from "@/components/Canvas/MobileWorkflowTreeMore.vue";
 import MobileWorkflowTreeNodesTab from "@/components/Canvas/MobileWorkflowTreeNodesTab.vue";
 import DebugPanel from "@/components/Panels/DebugPanel.vue";
-import { NODE_DEFINITIONS } from "@/types/node";
+import { resolveNodeDisplay } from "@/lib/nodeDisplay";
 import { useThemeStore } from "@/stores/theme";
 import { useWorkflowStore } from "@/stores/workflow";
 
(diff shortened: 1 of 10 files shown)


mbakgun:
Thanks @roonakyadav, this is close 🚀 

The resolver correctly uses `NODE_DEFINITIONS` as its source of truth, so new node types need no resolver-specific registration.

Three items remain before merge:

1. **`MobileNodeExecutionDetail.vue` was missed.** It still accesses `NODE_DEFINITIONS[props.node.type].label` directly. Opening the execution detail for an unknown node still crashes with `Cannot read properties of undefined (reading 'label')`. Please route its label, icon, color, and tile behavior through the resolver.

2. **The connection sheet changed existing behavior.** Its secondary line previously showed the node type, such as `HTTP`, but now shows a URL or message summary. Please add a `typeLabel` field to the resolver and preserve the original display.

3. **The mobile surface tests check source text rather than behavior.** These checks do not exercise runtime rendering, and the green suite still misses the execution detail crash. Please add SSR/component tests for the tree, connection sheet, and execution detail. `vue/server-renderer` with `createSSRApp` is sufficient and requires no new dependency.

Also, please align the new imports with the `AGENTS.md` order: Vue, external libraries, internal types, then internal code.

The rest looks solid. Thanks for working through these remaining cases. 🙏

roonakyadav:
Hey @mbakgun 
Addressed all three review items:

routed MobileNodeExecutionDetail through the shared resolver
restored the connection-sheet typeLabel behavior
replaced source-level checks with SSR/component coverage, including MobileWorkflowTreeNodesTab

Also cleaned up the test environment by mocking the theme store instead of mutating browser globals.

Validation: 162 tests passing, typecheck and lint clean.

mbakgun:
Thank you so much !

PR #541, opened 2026-09-08
fix: handle quotes and newlines in arithmetic string expressions

roonakyadav:
## Summary

Fixes an issue where runtime strings containing double quotes or newlines were incorrectly embedded into arithmetic expressions.

For example:

```text
$userInput.body.text + "!"
```

with:

```
He said "hello"
```

previously produced malformed expression text instead of:

```
He said "hello"!
```

The root cause was `resolve_arithmetic_expression()` wrapping runtime strings with `f'"{result}"'` before AST evaluation. This change uses `repr(result)` instead, matching the existing safe handling in the related expression evaluation paths.

## Changes

* Replace unsafe runtime string quoting with `repr(result)` in both affected branches of `resolve_arithmetic_expression()`.
* Add regression coverage for:
  * quoted runtime strings
  * multiline runtime strings
  * normal string concatenation

## Validation

* 178 expression evaluator tests passed.
* 113 related expression/execution tests passed.
* Verified with a real local Heym instance using PostgreSQL, Uvicorn, and the workflow execution HTTP API.
* Verified plain text, quotes, newlines, JSON-like text, backslashes, Unicode, and a dynamically generated runtime value.
* All real workflow executions returned the expected `input + "!"` result with `success` status.

Discussion: #539

### DIFF
--- a/backend/app/services/workflow_executor.py
+++ b/backend/app/services/workflow_executor.py
@@ -6236,14 +6236,14 @@ class WorkflowExecutor:
                 if result is None:
                     return "None"
                 if isinstance(result, str):
-                    return f'"{result}"'
+                    return repr(result)
                 return str(result)
             except Exception as e:
                 result = self._fallback_after_eval_error(e, expr, combined)
                 if result is None:
                     return "None"
                 if isinstance(result, str):
-                    return f'"{result}"'
+                    return repr(result)
                 return str(result)
 
         processed = self._replace_expressions(expression, replace_dollar_ref)
--- a/backend/tests/test_expression_evaluator_service.py
+++ b/backend/tests/test_expression_evaluator_service.py
@@ -975,6 +975,8 @@ class TestExpressionEvaluatorServiceEvaluate(unittest.TestCase):
             ("Hello world", "Hello world!"),
             ('He said "hello"', 'He said "hello"!'),
             ("Line 1\nLine 2", "Line 1\nLine 2!"),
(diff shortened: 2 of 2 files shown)


roonakyadav:
Addressed the issue reported in #539.

- Replaced unsafe runtime string quoting with `repr()` in both affected paths.
- Added regression coverage for quoted and multiline runtime strings.
- Verified with a real local Heym instance using PostgreSQL, Uvicorn, and the workflow execution API.

All relevant tests and CI checks are passing.

ckakgun:
This looks good @roonakyadav , thank you ! I applied it locally and confirmed the fix.

- Your case now returns `He said "hello"!` on a real run.

One request before I merge. Your tests cover quotes, newlines and plain strings, but
not the two cases you mention verifying by hand. Could you add them to the same
`cases` list?

```python
("C:\\next", "C:\\next!"),
('" + upper("pwn") + "', '" + upper("pwn") + "!'),
```

The first one used to rewrite backslash sequences silently, so `C:\next` came back
with a real newline in it. The second one used to evaluate the input as expression
syntax and return `PWN`. Your change fixes both, they are just not locked in yet.

Thanks for the clean patch and the clear report 💯

roonakyadav:
Thanks for the review. I’ve added the two requested regression cases for backslash handling and expression-like runtime input.

All checks are now passing, including the full expression evaluator test suite. The PR is ready for merge.

ckakgun:
Thank you so much Ronak 🌸

PR #543, opened 2026-09-09
fix: acknowledge offloaded RabbitMQ runs

roonakyadav:
## Summary

Fixes the RabbitMQ acknowledgement path for offloaded workflow executions when cluster execution is enabled.

When `_handle_message()` receives an offloaded result with `history_written=True`, it previously logged the result and returned before reaching the normal RabbitMQ ACK/NACK handling.

This left the original RabbitMQ delivery unacknowledged.

## Changes

- ACK offloaded runs with `status="success"`
- ACK offloaded runs with `status="pending"`
- NACK offloaded errors with `requeue=False`
- Preserve existing offloaded history persistence/deduplication behavior
- Add regression coverage for offloaded success, pending, error, timeout, and retired-run outcomes
- Add regression coverage confirming local execution paths retain their existing ACK/NACK and history behavior

## Validation

- RabbitMQ consumer tests: 8 passed
- Related trigger/cluster tests: 77 passed
- Ruff checks passed
- Ruff format check passed
- `git diff --check` passed

This implements the acknowledgement semantics confirmed by the maintainers in Discussion #542.

### DIFF
--- a/backend/app/services/rabbitmq_consumer.py
+++ b/backend/app/services/rabbitmq_consumer.py
@@ -376,6 +376,10 @@ class RabbitMQConsumerManager:
                     log_offloaded_run(
                         logger, workflow_id=workflow.id, trigger="RabbitMQ trigger", result=result
                     )
+                    if getattr(result, "status", None) in {"success", "pending"}:
+                        await message.ack()
+                    else:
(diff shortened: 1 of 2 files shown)


ckakgun:
Thank you Ronak 🚀

PR #545, opened 2026-09-10
fix(execution): preserve sub-workflow input types

roonakyadav:
## Summary

Fixes structured input type loss when an `execute` node passes `$...` expressions to a sub-workflow.

Previously, `executeInputMappings` and `executeInput` resolved `$...` expressions without `preserve_type=True`. This caused dictionary and list values to be serialized to JSON strings before reaching the child workflow.

For example:

- `dict` values arrived as JSON strings instead of dictionaries
- `list` values arrived as JSON strings instead of lists
- nested property access failed
- list indexing operated on the serialized string
- child loops iterated over the string instead of the original array

The maintainer confirmed this behavior is unintended in Discussion #544 and requested a narrowly scoped PR.

## Changes

- Preserve native types when resolving `$...` expressions in `executeInputMappings`
- Preserve native types when resolving `$...` expressions in `executeInput`
- Preserve the existing handling of normal text templates and scalar values
- Add regression coverage for:
  - dictionary input and nested access
  - list input and loop/indexing
  - scalar types
  - dictionary passed through `executeInput`
  - whole-object text rendering

## Validation

- ExecuteNode tests: 19 passed
- Related expression/output tests: 227 passed
- Ruff checks passed
- Ruff format check passed
- `git diff --check` passed

### DIFF
--- a/backend/app/services/node_execution/nodes/execute_node.py
+++ b/backend/app/services/node_execution/nodes/execute_node.py
@@ -74,9 +74,9 @@ def execute(ctx: NodeExecutionContext) -> object:
             key = mapping.get("key", "")
             value_template = mapping.get("value", "")
             if value_template:
-                if isinstance(value_template, str) and value_template.strip().startswith("$"):
+                if isinstance(value_template, str) and value_template.startswith("$"):
                     resolved_value = self.resolve_expression(
-                        value_template.strip(), inputs, node_id, preserve_type=True
+                        value_template, inputs, node_id, preserve_type=True
                     )
                 else:
                     resolved_value = self.evaluate_message_template(value_template, inputs, node_id)
@@ -84,11 +84,9 @@ def execute(ctx: NodeExecutionContext) -> object:
             else:
                 execute_inputs[key] = ""
     elif execute_input_template:
-        if isinstance(execute_input_template, str) and execute_input_template.strip().startswith(
-            "$"
-        ):
+        if isinstance(execute_input_template, str) and execute_input_template.startswith("$"):
             transformed_input = self.resolve_expression(
-                execute_input_template.strip(), inputs, node_id, preserve_type=True
+                execute_input_template, inputs, node_id, preserve_type=True
             )
         else:
             transformed_input = self.evaluate_message_template(
--- a/backend/app/services/node_execution/nodes/execute_node.py
+++ b/backend/app/services/node_execution/nodes/execute_node.py
@@ -74,16 +74,22 @@ def execute(ctx: NodeExecutionContext) -> object:
             key = mapping.get("key", "")
             value_template = mapping.get("value", "")
             if value_template:
-                if value_template.startswith("$"):
-                    resolved_value = self.resolve_expression(value_template, inputs, node_id)
+                if isinstance(value_template, str) and value_template.strip().startswith("$"):
+                    resolved_value = self.resolve_expression(
+                        value_template.strip(), inputs, node_id, preserve_type=True
+                    )
                 else:
                     resolved_value = self.evaluate_message_template(value_template, inputs, node_id)
                 execute_inputs[key] = resolved_value
             else:
                 execute_inputs[key] = ""
     elif execute_input_template:
-        if execute_input_template.startswith("$"):
-            transformed_input = self.resolve_expression(execute_input_template, inputs, node_id)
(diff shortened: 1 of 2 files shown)


mbakgun:
Thanks for the quick turnaround. The core fix is right and the tests cover the real signal, so this is close. One thing is blocking.

Along with `preserve_type=True` the change also adds `.strip()` before the `$` check. That reroutes values with leading whitespace away from the template path into the single expression path, which changes behavior for templates that work today.

Verified on a real parent to child run with `title = "Report"`, `count = 42`, `name = "Alice"`, `city = "Paris"`:

| mapping value | main | this PR |
|---|---|---|
| `Report: $parentInput.title` | `Report: Report` | `Report: Report` |
| ` $parentInput.title suffix` | ` Report suffix` | `null` |
| ` $parentInput.title / $parentInput.count` | ` Report / 42` | `null` |
| `  $parentInput.name and $parentInput.city  ` | `  Alice and Paris  ` | `Paris` |

The last row is the one to watch. When the stripped text happens to parse as valid Python it does not fail loudly, it returns a wrong value that looks reasonable. That is harder to spot than a null. Both `executeInputMappings` and `executeInput` are affected.

Could you drop the `.strip()` calls and keep just `preserve_type=True`? I ran that variant locally and it fixes dicts and lists exactly as intended while leaving all four rows above unchanged. The `isinstance` guard is fine to keep, it makes no observable difference either way.

You will also need to change the value in `test_execute_input_mappings_preserves_dict_and_nested_access` from `"  $parentInput.user  "` to `"$parentInput.user"`, and it would be good to add one of the mixed template rows above as a regression test.

Smaller note, not blocking. The loop test asserts `firstItem` and `loopTotal == 3`, which already proves the list arrived as a list. A child loop wired back into the loop handle would also prove the three iterations actually run.

Everything else looks good. CI is green, lint and format pass, and the five new tests pass here too. Happy to merge once the strip is out.

roonakyadav:
Thanks for the review. I’ve addressed the requested changes:

- Removed `.strip()` from both ExecuteNode expression paths.
- Updated the existing regression input to use the exact expression.
- Added coverage for mixed templates such as `$parentInput.name and $parentInput.city`.
- Re-ran the test suite successfully, including `test_execute_node_type_preservation.py`.

The updated commit is now pushed to the PR branch.

mbakgun:
Thanks for iterating. The mixed templates from the previous review now work, but two regressions remain. I verified both on real parent-to-child runs against `7fe5023`, in both `executeInputMappings` and `executeInput`.

First, the new `_is_single_dollar_expression` gate routes previously working arithmetic and comparison expressions into the text-template path:

| input value | main | this PR |
|---|---|---|
| `$parentInput.count + $parentInput.other` | `50` (int) | `"42 + 8"` (str) |
| `$parentInput.count > $parentInput.other` | `true` (bool) | `"42 > 8"` (str) |
| `$42` | `42` (int) | `"$42"` (str) |

Second, the helper trims its input when checking, but `resolve_expression` receives the untrimmed string. For a pure expression with leading whitespace, the resolver returns the source text unchanged:

| input value | main | this PR |
|---|---|---|
| `  $parentInput.title  ` | `"  Report  "` | `"  $parentInput.title  "` |
| ` $parentInput.count` | `" 42"` | `" $parentInput.count"` |

Passing `.strip()` into the resolver makes the expression resolve again, but it still does not match main. The surrounding whitespace is dropped, so `  $parentInput.title  ` returns `"Report"` rather than `"  Report  "`, and ` $parentInput.count` returns an int rather than `" 42"`. It also leaves the arithmetic, comparison, and numeric-literal regressions untouched.

Could we restore main's routing and limit this PR to type preservation? Keep the original `startswith("$")` checks and add only `preserve_type=True` to the two `resolve_expression` calls.

That preserves the existing distinction: inputs starting directly with `$` use the expression resolver, while inputs with leading whitespace use the text-template path. Please add regression coverage for the cases above in both input paths, alongside the dict/list tests.

This would leave the `$parentInput.name and $parentInput.city` mixed-template improvement for a separate PR. It is useful, but changing expression-versus-template routing needs broader compatibility coverage than this type-preservation fix.

Keeping this PR focused on preserving native input types should make it easier to validate and merge.

mbakgun (approved):
Thank you @roonakyadav ⭐️

PR #548, opened 2026-09-11
fix: support boolean ordering in workflow expressions

roonakyadav:
## Summary

Fixes boolean ordering for `DotBool` values in workflow expressions.

Boolean values coming from workflow context are wrapped in `DotBool`, but `DotBool` was missing the rich comparison operators required for `<`, `<=`, `>`, and `>=`.

This caused boolean comparisons to fail and also broke operations that rely on ordering, such as sorting and filtering.

## Changes

- Added `__lt__`, `__le__`, `__gt__`, and `__ge__` to `DotBool`.
- Added regression coverage for:
  - boolean comparisons
  - sorting boolean arrays
  - sorting objects by boolean fields
  - filtering using boolean ordering
- Covered both workflow execution and expression preview paths.

## Validation

- 3,996 backend tests passed
- Ruff check passed
- Ruff format check passed
- Frontend lint and typecheck passed
- File line-limit check passed

## Related

Closes #547

### DIFF
--- a/backend/app/services/workflow_executor.py
+++ b/backend/app/services/workflow_executor.py
@@ -1145,6 +1145,26 @@ class DotBool:
             return self._value == other._value
         return self._value == other
 
+    def __lt__(self, other: object) -> bool:
+        if isinstance(other, DotBool):
+            return self._value < other._value
+        return self._value < other  # type: ignore[operator]
+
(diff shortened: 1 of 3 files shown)


mbakgun:
Thank you so much 🌸

PR #562, opened 2026-09-17
fix: hand off cluster execution ownership to workers

roonakyadav:
## Summary

Fixes the cluster execution ownership race identified in Discussion #561.

Before this change, both the dispatcher and the claiming worker could register the same execution in `ActiveExecutionRegistry`. This allowed the dispatcher to continue heartbeating an execution after it had been offloaded to a worker.

This could cause:

- worker crashes to evade orphan recovery because the dispatcher kept the heartbeat fresh
- dispatcher shutdowns to incorrectly mark worker-owned executions as orphaned, allowing duplicate execution

## Changes

- Coordinate dispatcher → worker ownership handoff in `dispatch_workflow()`
- Relinquish dispatcher-local execution ownership after successful offload
- Prevent dispatcher cleanup from deleting a worker-owned execution
- Preserve worker ownership and progress against delayed registry writes
- Preserve cancellation state across worker registration
- Prevent queued / `waiting_for_main` executions from being treated as crashed orphans
- Handle cancellation before worker execution begins
- Add focused regression coverage for ownership handoff, cleanup, cancellation, delayed writes, and recovery

## Validation

- 105 affected backend tests passed
- `ruff check` passed
- `ruff format --check` passed
- `./check.sh` completed successfully
- The single failing test in the full runner was reproduced on a clean `5e393f0` baseline and is unrelated to this change

Related discussion: #561


mbakgun:
Thank you for working on this and addressing the concerns from the discussion. The direction is right, and several parts are indeed helpful 🌸  : keeping `cancel_requested_at` out of the conflict update, guarding the queued and `waiting_for_main` states in both recovery sweeps, and keeping queued runs visible. We did find several lifecycle issues that need to be resolved before merging so the fix does not introduce new regressions.

The main concern is that `_RELINQUISHED_EXECUTIONS` tracks execution IDs without distinguishing the dispatcher's registration from the worker's. Relinquishing ownership and completing an execution need to remain separate operations.

1. **The dispatcher can remove the worker's registration.**  
   `dispatch_workflow` reads the handle after awaiting enqueue and notification. A worker in the same process can register during those awaits, so the dispatcher picks up and relinquishes the worker's handle. The execution then loses registry heartbeats and cancellation tracking. Please make sure the handoff refers specifically to the dispatcher's original registration.

2. **The worker's finish command can be discarded.**  
   In the same process, `complete_execution` can queue a finish command while the ID is still marked relinquished. The registry drops it, and the dispatcher's later cleanup also skips deletion. A completed execution remains active and can be selected for recovery. Worker completion must remain effective regardless of when dispatcher cleanup runs.

3. **Delayed start writes can still reclaim ownership or recreate completed executions.**  
   `_apply_command` uses the drain time as `heartbeat_at`, so an old command can appear newer and pass the timestamp guard. It can also recreate a row after the worker has deleted it. Please ensure pending and in-flight writes from an earlier registration cannot overwrite the current owner or resurrect a finished execution. Changing the timestamp alone would not cover both cases.

4. **Expired queued runs can become recovery candidates.**  
   The recovery filters exclude `queued` and `waiting_for_main`, but that protection ends when the queue marks a run `skipped_late`. With dispatcher cleanup suppressed and no worker to finish it, the active row remains and can trigger execution of a deliberately skipped run. Terminal queue states need a defined cleanup path.

There are two additional issues to address:

- Relinquished IDs remain in memory for callers that do not call `clear_execution`, including dashboard and error workflow dispatches.
- The new cancellation path writes execution history without updating the analytics snapshot, unlike the in-process path.

Please add behavioral regression tests for these cases, especially both same-process orderings: worker registration before dispatcher relinquishment, and worker completion followed by registry drain before dispatcher cleanup. Delayed writes should also be tested after worker completion, along with queue expiry.

The existing `assertFalse(cleared)` assertion does not validate cleanup because `clear_execution` returns `None`. Please assert the resulting ownership and database state instead. The session test doubles should also support the real `AsyncSession` interface without requiring `hasattr` or `inspect.isawaitable` guards in production code.

Please preserve existing in-process cancellation and recovery behavior throughout. No backfill migration for historical stale rows is needed for this fix.

Thanks again. Happy to review the next revision 🙏

roonakyadav:
Thanks for the thorough review and guidance @mbakgun! 🙏

I have reworked the handoff architecture to address all points from the review:

1. **Dispatcher removing worker's registration:** Replaced the global `_RELINQUISHED_EXECUTIONS` ID set with instance-level handles tracking unique `registration_token: UUID`s. Dispatcher handles are captured before any async gaps, and `relinquish_execution()` will not evict a same-process worker handle that registered concurrently.
2. **Worker finish command dropped:** Removed the check in `_drain_commands` that skipped finish commands for relinquished IDs. Relinquishing drops only commands matching the dispatcher's specific registration token; the worker's finish command proceeds to delete the active row.
3. **Delayed start writes:** Commands now record `enqueued_at` timestamps used by the upsert `CASE` statement so older writes cannot overwrite a newer worker heartbeat. Added an anti-resurrection guard in `_apply_command` that checks `EXISTS` on `ExecutionHistory` and terminal `WorkflowRunQueue` states, discarding start commands for completed runs.
4. **Expired queued runs:** `expire_late_rows()` now atomically deletes `ActiveWorkflowExecution` rows when retiring runs to `skipped_late`. Orphan sweeps and stale cleanup also explicitly exclude all terminal queue states (`done`, `failed`, `skipped_late`).
5. **Memory leak:** Eliminated `_RELINQUISHED_EXECUTIONS`. Relinquishment state lives solely on the handle instance and is garbage collected normally.
6. **Analytics snapshot:** Worker cancellation in `RunQueueWorker._execute_claimed` now invokes `upsert_workflow_analytics_snapshot(..., status="cancelled")` to match in-process cancellation behavior.
7. **Assertions, Mocks & Defensive Cleanup:** 
   - `clear_execution()` and `complete_execution()` now return boolean statuses and tests assert state/ownership rather than mock calls.
   - Removed `hasattr`/`isawaitable` production guards; test doubles now adhere to the standard `AsyncSession` protocol.
   - Wrapped the offload enqueue path in `dispatch_workflow` to defensively clean up locally registered handles if `enqueue()` fails.
   - Added an unmocked integration test (`test_same_process_worker_completion_drains_finish_before_dispatcher_cleanup`) proving worker completion queues a finish command that drains to the database and issues `DELETE FROM active_workflow_executions` before dispatcher cleanup runs.

All 144 affected backend unit tests and linting/formatting checks are passing. Ready for your review!

mbakgun:
Thanks for the update 🙏 The registration tokens, same-process cleanup fixes, and analytics changes address several of the earlier concerns.

Four lifecycle gaps still need attention before merging:

1. An in-flight dispatcher heartbeat can overwrite worker ownership after handoff, because the database update still matches only `execution_id`.
2. Recovery can claim a long-waiting run after `claim_next` marks it `claimed`, but before the worker establishes a fresh heartbeat.
3. If the dispatcher's start command has not been flushed, relinquishing removes it and can leave a queued run without an active record for visibility or cancellation.
4. If history is written but active-row cleanup is interrupted, the new recovery exclusions leave stale records without an automatic cleanup path. The existing cleanup function is still not scheduled.

Please add behavioral regression tests for these orderings, ideally against PostgreSQL, and preserve existing in-process cancellation and recovery behavior. Cleanup should distinguish completed leftovers from recoverable work; simply removing the history exclusion or scheduling the broad stale-row deletion would not safely cover that distinction.

Thanks again for working through this. Happy to review the next revision.

roonakyadav:
Hi @mbakgun,

I’ve finished the follow-up changes based on your review feedback and pushed the latest updates to PR #562.

The main lifecycle fix is still centered around making execution ownership explicit during the dispatcher → worker handoff.

### What changed

- Reworked execution registration to use a unique `registration_token` per handle instead of relying on execution IDs alone.
- Added explicit `relinquished` state to dispatcher-owned handles.
- Dispatcher now captures its exact handle before crossing async boundaries and relinquishes that specific handle immediately after successful queue handoff.
- Worker registration gets its own ownership token, so dispatcher and worker heartbeats cannot be confused.
- Heartbeat synchronization is now scoped to the exact `worker_id` belonging to the handle, preventing an old dispatcher heartbeat from overwriting worker ownership.
- Delayed registry `start` commands retain their original enqueue timestamp and can no longer resurrect executions that have already reached a terminal state.
- Relinquishing a dispatcher handle removes only its pending `finish` commands while preserving the required `start` command for the worker lifecycle.
- `complete_execution()` / `clear_execution()` now operate on the exact handle where applicable rather than blindly acting on an execution ID.
- Added cleanup for active rows whose execution history has already been committed.
- `skipped_late` queue entries now clean up their active execution rows atomically.
- Queued cancellation is preserved through worker registration instead of being reset by the worker claim path.
- The in-process execution path remains unchanged in behavior; relinquishment only occurs for the cluster/offloaded path.
- Updated the relevant trigger/API callers so their cleanup uses the originating execution handle.

### Regression coverage

I added and updated focused tests covering:

- dispatcher/worker ownership handoff
- in-flight dispatcher heartbeat races
- delayed start-command races
- token-scoped handle lifecycle
- queued cancellation
- recently claimed execution protection from recovery
- terminal execution exclusion
- interrupted active-row cleanup
- late queue-row expiry
- worker-side cancellation/history handling

Most importantly, the same-process heartbeat race is now tested against a real PostgreSQL database rather than only mocks/structural assertions.

### Validation

The affected test suites currently pass:

- `test_active_execution_registry_resilience.py` — 28 passed
- `test_cluster_dispatch.py` — 32 passed
- `test_cluster_run_queue.py` — 12 passed
- `test_execution_cancellation.py` — 26 passed
- `test_execution_recovery.py` — 19 passed
- `test_rabbitmq_consumer.py` — 33 passed

Total: **150 passed, 0 failed, 0 skipped**

The real PostgreSQL concurrency regression test also passes.

I also fixed the CI environment issue exposed by the new PostgreSQL-backed test. The normal `check` job now provisions PostgreSQL, runs the migrations, and provides the test database through `DATABASE_URL`.

The latest commit is:

`6bfbdae` — `ci: provide postgres for backend tests`

GitHub Actions is now green for the PR.

Please take another look at the implementation and the ownership/race handling when you get a chance. I specifically tried to keep the fix limited to the lifecycle/ownership problem you identified while preserving the existing in-process behavior.

mbakgun:
Thanks for the follow-up and the PostgreSQL regression test 🙏 The changes address several earlier concerns, but two lifecycle gaps remain:

1. If the dispatcher writes a newer heartbeat after worker registration but before the worker's start command drains, the timestamp guard keeps dispatcher ownership. Subsequent worker heartbeats fail the ownership check, and the fallback skips reinsertion because the row exists. A live execution can therefore become eligible for recovery. Please cover the actual start-command path; the current PostgreSQL test assigns worker ownership directly with SQL.

2. Preserving the start command still leaves a gap before it reaches the database. After relinquishment, a queued run can have neither a local handle nor an active row, so cancellation returns not found. Visibility and durable cancellation need to remain available throughout the handoff.

The PostgreSQL test should also be isolated from existing development data and clean up all fixtures, including any user it creates. CI provisioning is helpful, but the local test path now requires a running, migrated PostgreSQL database. Please provide a clear setup path in `check.sh` and record the prerequisite in `AGENTS.md`, while keeping this regression test mandatory in CI.

Please address these before merging and preserve existing in-process cancellation and recovery behavior. Thanks again for working through this.

roonakyadav:
Hi @mbakgun,

I’ve pushed the final set of changes for the execution ownership / handoff issues to PR #562 in commit `76b7ccb`.

The implementation now addresses the lifecycle races from your previous review:

### Ownership / heartbeat

- Dispatcher and worker registrations use distinct registration tokens.
- Dispatcher heartbeat updates are scoped to the exact owning handle.
- Worker ownership cannot be overwritten by an in-flight dispatcher heartbeat.
- Worker start commands are preserved even after the queue transitions to `STATUS_CLAIMED`.
- Finish commands are now owner-scoped, so a stale dispatcher finish cannot delete a worker-owned `ActiveWorkflowExecution`.
- Terminal-state checks prevent delayed registry writes from resurrecting completed executions.

### Cancellation during handoff

The cancellation path now explicitly handles the window where:

`enqueue → dispatcher relinquish → worker registration`

and the active execution row may temporarily not exist.

Cancellation now uses the queue row as the durable source during that window, persists the cancellation state, and the worker checks the persisted cancellation before invoking the workflow executor.

The real PostgreSQL regression test verifies that cancellation during this handoff succeeds and the worker does not execute the cancelled workflow.

### Regression coverage

I added deterministic regression coverage for:

- in-flight dispatcher heartbeat vs worker ownership
- cancellation during dispatcher → worker handoff
- stale dispatcher finish vs worker-owned active row
- worker start command after queue claim
- delayed start-command handling
- ownership-token lifecycle behavior

These tests run against real PostgreSQL where the invariant depends on database concurrency.

### Validation

The PR-specific validation is clean:

- `test_active_execution_registry_resilience.py`: 32/32 passed
- Core cluster/execution suites: 124/124 passed
- Real PostgreSQL ownership/cancellation race tests: passed
- Stale dispatcher finish regression: passed
- Worker start ownership regression: passed
- Ruff / formatting: passed
- Frontend lint/typecheck: passed
- Line-limit and diff checks: passed

The full backend suite completed with 4,113 tests and 4,112 passing. The only failures were two unrelated timing-sensitive tests:

- `test_execute_do_not_wait_stream_release.py`
- `test_execute_do_not_wait_pool_isolation.py`

I also reproduced those failures on a clean baseline worktree at the pushed base commit `6bfbdae` using the same parallel test setup. Both passed when run individually on both baseline and PR. Therefore I deliberately left those unrelated tests, timing thresholds, and global test concurrency unchanged rather than masking the failures inside this PR.

The latest commit is:

`76b7ccb` — `fix(cluster): harden execution handoff and cancellation`

Please review the latest lifecycle changes when you get a chance.

mbakgun:
Thanks for the update and the fixture cleanup. A few changes are still needed before merging:

- A delayed dispatcher start command can reclaim ownership after another process has registered the worker. A local handle does not establish queue ownership. Please tie the handoff to the claiming worker so the dispatcher cannot take ownership back.
- Queued cancellation writes `cancelled` history, but setting the queue's `error` field makes the waiting dispatcher receive an error instead. Please preserve the cancelled result and update analytics in this path.
- Keep queued runs visible between relinquishment and the first active-row write.
- Align the PostgreSQL setup instructions with what `check.sh` actually does, and avoid printing `DATABASE_URL`, which can contain credentials.

Please preserve existing in-process cancellation and recovery behavior. Thanks again for working through this.

mbakgun (changes requested):
Once you've applied the changes here, I'm happy to proceed with this PR. Thanks again.

PR #564, opened 2026-09-18
fix(hitl): claim review requests atomically to prevent double resume

### LINKED ISSUE #563: HITL decisions can be double-submitted concurrently, resuming the paused workflow twice
## Summary

`POST /api/hitl/{token}/decision` performs a read-check-write on `hitl_requests.status` without any row lock or conditional update. Two concurrent decision submissions for the same review request both pass the `pending` check, both commit, and both schedule the background resume — so a paused workflow can be resumed twice with duplicate side effects, and a second submission silently overwrites the first reviewer's decision.

## Affected code

- `backend/app/api/hitl.py` — `submit_hitl_decision` (the public token-authenticated decision endpoint)
- `backend/app/api/ai_assistant.py` — the `resolve HITL review` agent-tool path (`asyncio.create_task(resume_hitl_request_in_background(...))` after the same non-atomic read-check-write)

Tested against commit `5e393f09a3b98c0d110a16989bb5da23afb3657d` (v0.0.110).

## Steps to reproduce

1. Start the stack locally (Postgres + backend).
2. Insert a `pending` HITL request row (any workflow/execution parent rows are enough; the decision endpoint only needs the token):
   ```sql
   INSERT INTO hitl_requests (id, workflow_id, execution_history_id, public_token,
     workflow_name, agent_node_id, agent_label, summary, original_draft_text,
     original_agent_output, resolved_output, execution_snapshot, status, expires_at)
   VALUES (gen_random_uuid(), <workflow_id>, <history_id>, 'race-test-token',
     'wf', 'n1', 'agent', '', '', '{}', '{}', '{}', 'pending', now() + interval '7 days');
   ```
3. Fire two decisions **concurrently** (e.g. `asyncio.gather` with two `httpx` clients, or two simultaneous `curl` calls):
   ```bash
   curl -X POST http://localhost:10105/api/hitl/race-test-token/decision \
        -H 'Content-Type: application/json' -d '{"action": "accept"}' &
   curl -X POST http://localhost:10105/api/hitl/race-test-token/decision \
        -H 'Content-Type: application/json' -d '{"action": "refuse"}' &
   wait
   ```
4. **Expected** (serialized requests): first returns `202`, second returns `409 Conflict` ("Review request has already been resolved").
   **Actual** (concurrent requests): **both return `202`** — both transactions read `status == "pending"` before either commits.

Verified live against a local stack (Postgres 16 + backend v0.0.110, commit `5e393f0`):

```text
== concurrent submissions ==
{"action": "accept", "http_status": 202, "body": {"request_id": "…", "status": "resolved"}}
{"action": "refuse", "http_status": 202, "body": {"request_id": "…", "status": "resolved"}}
== serialized follow-up (control) ==
{"action": "accept", "http_status": 409, "body": {"detail": "Review request has already been resolved"}}
```

Database state afterwards: `status = resolved`, `decision = 'accept'` — the concurrent `refuse` reviewer received `202 Accepted` while their decision was silently discarded (last write wins). Both submissions also scheduled `resume_hitl_request_in_background`, and each instance passes the `status != "resolved"` guard since both writers set the same status — so the paused workflow is resumed **twice** with the owner's restored credential context.

## Why the second line of defense does not help

`resume_hitl_request_in_background` re-reads the request and bails out unless `status == "resolved"` with a decision (`hitl_service.py`). That guard cannot prevent a double resume here: **both** writers set the same `resolved` status, so both scheduled background tasks pass it and each calls `resume_workflow_execution` — the paused workflow runs twice, with the workflow owner's credential context restored from the snapshot. When the two decisions differ (e.g. `accept` vs `refuse`), the second commit silently overwrites the first reviewer's choice (last write wins), and the workflow is resumed with whichever payload landed last — while the first reviewer was told their decision was accepted.

## Impact

- The paused workflow is resumed **twice** under the owner's credentials: duplicate notification sends, duplicate HTTP/tool calls, duplicate sub-workflow executions, doubled LLM spend.
- A reviewer's decision can be silently overwritten by a concurrent submission and the workflow still resumed, so the audit trail (`hitlHistory`) no longer matches what actually executed.
- Analytics snapshots (`upsert_workflow_analytics_snapshot`) are updated for each resume.

This is the same family of state-machine race conditions already fixed in #217 / #219; the decision endpoint appears to be the remaining unguarded writer.

## Suggested fix

Make the claim atomic, e.g.:

```python
result = await db.execute(
    update(HITLRequest)
    .where(HITLRequest.id == hitl_request.id, HITLRequest.status == "pending")
    .values(decision=..., status="resolved", resolved_at=..., ...)
)
if result.rowcount == 0:
    raise HTTPException(409, "Review request has already been resolved")
```

or `SELECT ... FOR UPDATE` before the check. Applying the same guard to the `ai_assistant.py` agent-tool path would close both writers.

xiaodu55:
Closes #563.

## Problem

`submit_hitl_decision` (public `POST /api/hitl/{token}/decision`) and the `resolve_hitl_review_tool` assistant path used a read-check-write on `hitl_requests.status` with no row lock or conditional update. Two concurrent submissions both passed the `pending` check, both committed, and both scheduled `resume_hitl_request_in_background`: the paused workflow resumed twice with the owner's credential context, and one reviewer's decision silently overwrote the other's while both callers received `202`.

Verified live on v0.0.110 (`5e393f0`) against Postgres 16 — two concurrent submissions:

```text
{"action": "accept", "http_status": 202, "body": {"request_id": "…", "status": "resolved"}}
{"action": "refuse", "http_status": 202, "body": {"request_id": "…", "status": "resolved"}}
```

…with the losing decision silently discarded in the final row state.

## Fix

- New `claim_hitl_request_for_decision()` in `hitl_service.py`: a single conditional `UPDATE … WHERE id = :id AND status = 'pending' AND expires_at > now()`. The `status`/`expires_at` predicates live in the UPDATE itself, so exactly one concurrent caller wins and expiry is enforced atomically on the same statement.
- Both entry points (public decision endpoint, assistant tool path) claim before writing the decision fields and `resolved_output`. After a successful claim the row is exclusively ours, so mirroring the claimed state onto the session object and committing cannot race.
- A lost claim re-reads the row and surfaces the same errors a serialized caller would see: `409 Conflict` ("already been resolved") or `410 Gone` (expired). It never schedules a resume.

## Verification

After this change, same concurrent scenario:

```text
== concurrent submissions ==
{"action": "accept", "http_status": 202, "body": {"request_id": "…", "status": "resolved"}}
{"action": "refuse", "http_status": 409, "body": {"detail": "Review request has already been resolved"}}
== serialized follow-up (control) ==
{"action": "accept", "http_status": 409, "body": {"detail": "Review request has already been resolved"}}
```

Reproduced across multiple runs (winner alternates, loser always `409`, latency in milliseconds). The public endpoint and the assistant tool path were both exercised.

## Tests

New `tests/test_hitl_decision_atomic_claim.py` (mock style matching the existing HITL regression tests):

- winner path: claims once, schedules exactly one resume, returns `resolved`;
- lost claim on both entry points: `409`, no second resume scheduled;
- expiry mid-flight: `410`, no resume;
- the claim statement carries the `status`/`expires_at` predicates.

Full local run: 4068 passed. The handful of failing suites (`test_code_python_executor`, `test_codex_docker_wrapper`, `test_opencode_runner_service`, `test_skill_python_executor_*`, `test_execute_do_not_wait_*`) fail identically on a clean checkout of `main` on Windows (they exercise Docker/Unix-socket sandboxes) and are unrelated to this change.


mbakgun:
Thanks for turning this around so quickly 🙏

We verified the fix. Across repeated concurrent runs there is now exactly one 202 and one 409, the winner alternates, and the resolved row stays consistent. The conditional UPDATE is the right layer, and keeping the lock out of `get_hitl_request_by_token` was a good call.

One blocker before merge. The `await db.rollback()` in `refresh_hitl_request_after_lost_claim` rolls back the whole request session, not just the HITL row. The chat endpoints load `current_user` and the credential on that same session, so after a lost claim the next attribute access raises `MissingGreenlet` and the chat turn breaks. It also drops any uncommitted work from earlier tool calls in that turn. Refreshing just the one row, for example `populate_existing=True` on the re-read, would keep the expiry branch correct without the collateral.

Two smaller notes. `test_claim_predicates_include_status_and_expiry` asserts `"status" in sql`, but `status` also appears in the SET clause, so the test still passes if the `status == "pending"` predicate is removed. Asserting against `statement.whereclause` would close that. A test that a chat turn survives a lost claim would also help, since the mocked rowcount cannot reach that path.

The Codex follow-up endpoint has the same shape, but that one is pre-existing and does not need to ride along here.

xiaodu55:
Fixed in 7062f30 — thanks for catching the session blast radius, that would have broken chat turns on every lost claim.

- `refresh_hitl_request_after_lost_claim` no longer rolls back: it re-reads just the HITL row with `populate_existing=True`, so the chat user, credentials, and any uncommitted work from earlier tool calls in the turn are untouched. The expiry branch still maps to 410, resolved stays 409, and a vanished row maps to 404.
- `test_claim_predicates_include_status_and_expiry` now asserts the predicate columns against `statement.whereclause` ({id, status, expires_at}) instead of raw SQL.
- Added `RefreshHitlRequestAfterLostClaimTest` covering the no-rollback reload and its 409/410/404 mapping, and the lost-claim tests on both entry points now assert `db.rollback` is never called.

Agreed on the Codex follow-up endpoint staying out of this PR — happy to take it as a follow-up once this lands.

mbakgun (approved):
Thank you so much 🌸

